PR #2, opened 2017-03-23
ENH enable openmp on osx

### LINKED ISSUE #1: ENH handle OpenMP on linux and mac osx
So right now, the OpenMP support between linux and osx is spotty.
1. OSX has no support for OpenMP.
2. If the proper libraries are not in the linker path for OpenMP on linux, then the recipe will fail.

I think the easiest way to fix this would be to compile with conda's `gcc` for osx only and then include `libgcc` as a run dependence. 

See [this](https://github.com/conda-forge/staged-recipes/pull/2628) PR for osx and this [comment](https://github.com/conda-forge/staged-recipes/pull/2573#discussion_r107445878) from @jakirkham.

beckermr:
I don't expect to merge this yet, but starting work now.

Closes #1

### DIFF
--- a/recipe/Makefile.patch
+++ b/recipe/Makefile.patch
@@ -6,8 +6,8 @@
  ifeq ($(OS), Darwin)
 -export CC = $(if $(shell which gcc-5),gcc-5,clang-omp)
 -export CXX = $(if $(shell which g++-5),g++-5,clang-omp++)
-+export CC = clang
-+export CXX = clang++
++export CC = gcc
++export CXX = g++
  endif
  
  export LDFLAGS= -pthread -lm $(ADD_LDFLAGS) $(DMLC_LDFLAGS) $(PLUGIN_LDFLAGS)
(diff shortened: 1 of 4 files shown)


beckermr:
@jakirkham Thanks again for all the help! The various OpenMP related changes are collected here in this PR. Take a look when you have time. No rush!

beckermr:
@jakirkham It looks the various clang and openmp related recipes have been merged. Should we redo this recipe to enable openmp on OSX with those tools, or are we OK with this recipe as is? 

I rebuilt it and it still passes all of the tests.

jakirkham:
It's definitely an interesting approach to consider. The fact that the LLVM toolchain is supported through packages in `conda-forge` and is a reasonably recent version definitely makes it an attractive alternative to `gcc`. It's also a bit more friendly with UCRT on Windows, which is definitely a plus. I tried building one of my dependencies that uses OpenMP with the LLVM toolchain and was able to run all tests without issues. So it's certainly worth exploring.

beckermr:
OK. @jakirkham. I am closing this in favor of #4. Hopefully using clang will help with windows (#3) as you suggest!

PR #4, opened 2017-07-16
ENH OpenMP w/ clang for OSX

### LINKED ISSUE #1: ENH handle OpenMP on linux and mac osx

beckermr:
Closes #1 

Is an alternative to #2

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -15,7 +15,7 @@ source:
     - build-python.patch  # [osx]
 
 build:
-  number: 0
+  number: 1
   skip: True  # [win]
   script: python setup.py install --single-version-externally-managed --record record.txt
 
@@ -24,11 +24,13 @@ requirements:
     - python
     - setuptools
     - toolchain
+    - gcc  # [osx]
 
   run:
     - python
     - numpy
     - scipy
+    - libgcc
 
 test:
   requires:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -24,13 +24,12 @@ requirements:
     - python
     - setuptools
     - toolchain
-    - gcc  # [osx]
+    - openmp  # [osx]
 
   run:
     - python
     - numpy
     - scipy
-    - libgcc
 
 test:
   requires:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -24,12 +24,15 @@ requirements:
     - python
     - setuptools
     - toolchain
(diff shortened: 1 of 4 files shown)


jakirkham:
Will need to add `libcxx` as a dependency too.

beckermr:
Thanks! I got it to compile, but the linking is off. :(

beckermr:
Mmmmk. @jakirkham Take a look and let me know if you think this is good to go!

recipe/meta.yaml
```diff
@@ -24,11 +24,15 @@ requirements:
     - python
     - setuptools
     - toolchain
+    - libcxx  # [osx]
```
jakirkham: Would add `clangdev` for macOS above as that is going to be needed to build with `libcxx` and `openmp`. Though it is getting pulled in automatically ATM. That probably won't be the case going forward.

recipe/meta.yaml
```diff
@@ -24,11 +24,15 @@ requirements:
...
+    - openmp  # [osx]
```
jakirkham: Would pin these like so.

```yaml
     - libcxx 4.0.*  # [osx]
     - openmp 4.0.*  # [osx]
```

recipe/meta.yaml
```diff
@@ -24,11 +24,15 @@ requirements:
...
 
   run:
     - python
     - numpy
     - scipy
+    - libcxx  # [osx]
+    - openmp  # [osx]
```
jakirkham: Also would pin these like so.

```yaml
     - libcxx 4.0.*  # [osx]
     - openmp 4.0.*  # [osx]
```

jakirkham:
Sorry this fell off the radar. Have added a couple of comments above. Some of these were determined relatively recently like the pinnings.

beckermr:
I will make another or and merge.

gwerbin:
Does this work on Linux or just MacOS?

beckermr:
It should work on all systems except windows.

PR #7, opened 2018-01-16
Update requirements and rerender

ukaratay:


beckermr:
Thanks for the PR! Circle 2.0 is amazing. I assume you can merge. 

I think you need to increment the build number in the conda config yml before merging.

ukaratay:
@beckermr Increased the build number, thanks for catching that. Will you merge? I don't think I have the permission to do so.

aldanor:
Just wondering re: `4.0.*` -> `>= 4.0` change for `openmp` and `libcxx` on mac, is that intentional?

recipe/meta.yaml
```diff
@@ -24,15 +24,15 @@ requirements:
     - setuptools
     - toolchain
     - clangdev
-    - libcxx 4.0.*  # [osx]
-    - openmp 4.0.*  # [osx]
+    - libcxx >=4.0  # [osx]
+    - openmp >=4.0  # [osx]
```
isuruf: You should build with `4.0.*` and let the run requirements be `>=4.0`. Otherwise this package will be built with `5.0`, but at runtime 4.0 satisfies the constraints

beckermr: Good catch! Does package built with 4.0 run ok with 5.0?

ukaratay: @aldanor, yes, that was intentional. openmp and libcxx got updated to version 5 and that causes xgboost to downgrade when other packages are installed.

beckermr: I also do not understand the current status of these libs and what conda-forge wants to be standard.

ukaratay: @beckermr let me try

isuruf: > Does package built with 4.0 run ok with 5.0?

I've no idea. If you are not sure, then have `5.0.*` in both build an run requirements

ukaratay: @beckermr, @isuruf it does work when built with 4.0 and run with 5.0.

beckermr:
WTF travis?

ukaratay:
@beckermr, it looks like travis tests are gonna take a while to run: https://www.traviscistatus.com/incidents/mc9x2wmpnvhg

I tested it with py36 on macos. It builds and runs.

beckermr:
I'm ok to wait on the CI to be sure.

PR #18, opened 2018-11-23
Add python 3.7 support

jan-janssen:
It would be nice to add Python 3.7 support.


beckermr:
So conda forge is in the middle of upgrading everything to use the new compilers and an automated PR will be made for python 3.7 support at the same time. 

This pr has not yet been issued for this feedstock. My guess is that the coupling of the R package means not all of the dependencies are ready yet. 

I think we should wait for this automated PR.

PR #20, opened 2018-12-29
update to 0.81

bgruening:
Checklist
* [x] Used a fork of the feedstock to propose changes
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -19,7 +19,7 @@ source:
     - 0005-Add-ADD_-FLAGS-to-end-of-commandlines-osx.patch  # [osx]
 
 build:
-  number: 1001
+  number: 1000
   skip: true  # [win or linux32]
 
 requirements:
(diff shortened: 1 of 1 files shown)


beckermr:
I believe the build number needs to be set to 1000? We might want to update the PR instructions in conda-smithy.

bgruening:
@beckermr I can do this. I don't think we need to update the instructions as the build-number will go back to 0 as soon as the big GCC7 migration is done. But for now, I'm just curious if it still builds ... I fear that we need to update the patches :(

beckermr:
Yeah the patches...

These were inherited from the defaults build when I merged that in. We need to submit them upstream.

bgruening:
wow, that was easy :)

Issue #41, opened 2019-10-28
R ignoring openmp on osx


beckermr:
The R build is ignoring openmp on osx.

jakirkham:
I wonder if something is going on [in here]( https://github.com/dmlc/xgboost/blob/515f5f5c4779ff5361dcd796e22d55937e1048ea/R-package/configure#L1673-L1696 ).

jakirkham:
That said, does OpenMP actually get used in the R bindings? Or is it actually just used in `libxgboost`, which the bindings merely call? If it's the latter, then maybe the issue is that Linux is adding an extra linkage to R binding package that it shouldn't.

beckermr:
IDK much about the internals. I did notice in the configure script for the R build it says it is not finding openmp. 

Also, I don't see where the R build actually links to libxgboost. I am not confident it is actuall using the code there. Conda build certainly is not finding any linkages.

jakirkham:
Yeah there appears to be some [C binding code for the R package]( https://github.com/dmlc/xgboost/tree/v0.90/R-package/src ).

Now that I'm looking at it, it does appear to be using OpenMP.

jakirkham:
@RAMitchell, would you be able to advise us on how to proceed here?

RAMitchell:
I'm not familiar with osx or R. @hcho3/@trivialfis do you have any ideas?

hcho3:
There are two ways to build XGBoost-R package. 

1. Using `install.packages()`. This will download a tarball from CRAN and build XGBoost using GNU toolchains. Since Apple Clang doesn't support OpenMP out of box, we disable OpenMP for this use scenario.

2. Using CMake. The user can check out the XGBoost code using git and use CMake to build XGBoost with R binding. In this case, OpenMP will be enabled.

jakirkham:
Thanks for the feedback.

We are [using git to download the code]( https://github.com/conda-forge/xgboost-feedstock/blob/23eaf9a71f994dfac45bd69cd1b88a42a5238d86/recipe/meta.yaml#L11-L12 ), but it appears we are using [`r cmd install --build`]( https://github.com/conda-forge/xgboost-feedstock/blob/23eaf9a71f994dfac45bd69cd1b88a42a5238d86/recipe/install-r-xgboost.sh#L16 ) instead of CMake for the build of the R bindings. So it sounds like we should change that to CMake then.

[...]

PR #42, opened 2019-10-29
Build R packages with CMake

jakirkham:
Switches the build of the R bindings to use CMake. As part of this, the R bindings are now built as part of the overall build step. This should ensure that the packages are built with OpenMP support on macOS.

Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

Fixes https://github.com/conda-forge/xgboost-feedstock/issues/41

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -10,8 +10,12 @@ jobs:
   strategy:
     maxParallel: 8
     matrix:
-      linux_:
-        CONFIG: linux_
+      linux_r_base3.5.1:
+        CONFIG: linux_r_base3.5.1
+        UPLOAD_PACKAGES: True
+        DOCKER_IMAGE: condaforge/linux-anvil-comp7
+      linux_r_base3.6:
+        CONFIG: linux_r_base3.6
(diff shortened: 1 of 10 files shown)


beckermr:
It looks like the build is making either the shared lib or the R library. So I am running it twice.

beckermr:
We're getting extra R linkages in places that really shouldn't have them.

```
WARNING (libxgboost,bin/xgboost): $RPATH/libR.so not found in sysroot, is this binary repackaging? .. do you need to use install_name_tool/patchelf?
```

I think we should switch to running cmake in the subpackage build scripts. 

Also, if the R package is not going to link to libxgboost then I don't actually see a reason to have it around.

jakirkham:
> We're getting extra R linkages in places that really shouldn't have them.

That's concerning.

> I think we should switch to running cmake in the subpackage build scripts.

Yeah I agree with that. We could also look at patching the normal R build to use OpenMP if that is easier.

> Also, if the R package is not going to link to libxgboost then I don't actually see a reason to have it around.

Not sure if I'm completely understanding. The R bindings link to libxgboost. Though maybe you are saying something else here.

PR #50, opened 2020-06-30
Upgrade to 1.1.1

hcho3:
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

cc @jakirkham


hcho3:

recipe/install-py-xgboost.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 pushd ${SRC_DIR}/python-package
-  ${PYTHON} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv
+  ${PYTHON} setup.py install --single-version-externally-managed --record=record.txt
```
jakirkham, line 4: Could you please elaborate on why `pip` can't be used here now?

recipe/meta.yaml
```diff
@@ -9,15 +9,14 @@ source:
   # we have to use git urls here to get the submodules needed for the
   # build
   git_url: https://github.com/dmlc/xgboost
-  git_tag: 917b0a7b46954e9be36cbc430a1727bb093234bb
+  git_tag: d6d8be6519185d2e78ba7de5c68e6b7c9733b6f2
   patches:
     # xgboost patches
     - 0001-conda-Unbundle-libxgboost.-dll-dylib-so.patch
     - 0002-Fix-R-package-PKGROOT.patch
```
jakirkham, line 16: Do you know if it would be possible to upstream these into xgboost as well?

recipe/install-py-xgboost.sh
beckermr, line 4: Pls use pip.

beckermr: Or add a comment as to why pip doesn’t work.

hcho3: @jakirkham @beckermr The reason is similar to what happened in https://github.com/conda-forge/staged-recipes/pull/11926: `setup.py` pulls from `libxgboost.so` in the `build` directory, and somehow `pip install .` fails.

hcho3: I will open a new issue in XGBoost and see if `pip install .` can be made to work. Right now, it does not.

recipe/meta.yaml
hcho3: It would be nontrivial, since these files affect how XGBoost gets deployed to [CRAN](https://cran.r-project.org/web/packages/xgboost/index.html). To us (XGBoost project maintainers), the priority is to have XGBoost available on CRAN, since vast majority of users of R install packages from CRAN.

recipe/install-py-xgboost.sh
jakirkham: Thanks for the context. I think that is what [this patch]( https://github.com/conda-forge/xgboost-feedstock/blob/696ca06a4f429484a2c970b3c67c1f807481e7a9/recipe/0001-conda-Unbundle-libxgboost.-dll-dylib-so.patch ) is addressing. Does that address the problem sufficiently from your perspective or are there other things we should be doing?

recipe/meta.yaml
jakirkham: Meaning `0002-Fix-R-package-PKGROOT.patch`? Do you have thoughts on `0001-conda-Unbundle-libxgboost.-dll-dylib-so.patch`?

recipe/install-py-xgboost.sh
hcho3: @jakirkham The problem actually has to do with the recent refactoring of `setup.py` in https://github.com/dmlc/xgboost/pull/5271, so the patch does not apply. We will need to find a new way to make `pip` work with the new `setup.py`.

recipe/meta.yaml
hcho3: The first patch can be upstreamed. Let me do that by the next release (1.2.0).

recipe/install-py-xgboost.sh
jakirkham: Ah ok. Should we drop the patch here then?

recipe/meta.yaml
jakirkham: Sounds good. Is there an issue for tracking this?

recipe/install-py-xgboost.sh
hcho3: I actually modified the 0001 patch so that it only adds an entry to libpath.

recipe/0001-conda-Unbundle-libxgboost.-dll-dylib-so.patch
```diff
@@ -1,42 +1,15 @@
...
+index 65f7403e..4d044cf8 100644
 --- a/python-package/xgboost/libpath.py
 +++ b/python-package/xgboost/libpath.py
-@@ -24,6 +24,10 @@ def find_lib_path():
-                 os.path.join(curr_path, './lib/'),
-                 os.path.join(sys.prefix, 'xgboost')]
-     if sys.platform == 'win32':
+@@ -25,6 +25,10 @@ def find_lib_path():
+         # editable installation, no copying is performed.
+         os.path.join(curr_path, os.path.pardir, os.path.pardir, 'lib'),
```
ksangeek, line 7: PMFJI, When I had attempted to update for the recipe for a newer version, I too had noticed using `pip` did not work. I had also noticed that the `libxgboost.so` was getting copied into the python package, which I thought was best to avoid. Does this patch here does away with that problem?

hcho3: No, it does not. See https://github.com/dmlc/xgboost/issues/5386.

ksangeek: I was assuming that was caused by not using `pip`(my bad!).
Thanks for the pointer.

recipe/meta.yaml
hcho3: I opened https://github.com/conda-forge/xgboost-feedstock/issues/51

jakirkham:
Any other thoughts @beckermr @ksangeek? 🙂

jakirkham:
cc @dpryan79 (for awareness)

hcho3:
Can we merge this?

jakirkham:
Thanks @hcho3! 😀

PR #43, opened 2019-11-19
Attempt Windows build

xhochy:
Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #59, opened 2021-02-05
Attempt windows build 2

PrimozGodec:
Checklist
* [ ] Bumped the build number (if the version is unchanged)


PrimozGodec:
@conda-froge-admin, please rerender

PrimozGodec:

PrimozGodec:

wagnerpeer:
Thank you for the work, I had no time to do it myself and priority was pretty low. 

Looks like the OSX and Linux build are failing, because of missing dependencies, which needs to be build before. But I am not familiar with this. Has this something to do with Python 3.6?

The windows conda-build however fails, because it cannot download required code (maybe a patch?).
Found this, which seems pretty similar: https://help.appveyor.com/discussions/problems/3712-git-remote-https-seems-to-be-broken 

Also, the used task to install and configure conda environments is marked as deprecated.

I will have a look into all this, in the next week.

trivialfis:
Feel free to ping me if there's anything I can help. ;-)

PrimozGodec:
@wagnerpeer and @trivialfis thank you for offering help. I believe we can somehow make it work.

First about fail on Windows. Has anyone seen that before or have an idea what to do?
The error:
```
Cloning into bare repository '/d/bld/git_cache/github.com/dmlc/xgboost'...
error: remote-curl: usage: git remote-curl <remote> [<url>]
```
The only similar issue I can find is this one: https://stackoverflow.com/questions/48532045/git-clone-issue-in-msys2-shell

PrimozGodec:
@conda-forge/help-python, does anyone has an idea what is going on with the Windows build. The issue is highlighted in the previous comment. I would appreciate any help.

wagnerpeer:
To get the build running, we might need to reduce to a minimum working recipe. I would suggest to drop Python 3.6 and 3.9 to only try the most stable and widely supported Python versions at the moment. (see `.ci_support/win_64_.yaml`)

Also, I suspect issues with applying the patches. To get more information from the command executed, we could extend the build step by defining the following two variables in the `.azure-pipelines/azure-pipelines-win.yml` file:
```yaml
- step: |
    # ...
  displayName: Build recipe
  env:
    # ...
    GIT_CURL_VERBOSE: 1
    GIT_TRACE: 1
```

Another thing, for better readibility and traceability, I suggest to change the value of `git_tag` in the `recipe/meta.yaml` to the official tag name `v1.3.0`.

At last some questions:
* Do we need a bld.bat to build the library?
* Are we trying to build all packages (libxgboost, py-xgboost, py-xgboost-cpu, xgboost, _r-xgboost-mutex, r-xgboost and r-xgboost-cpu)?
  * Shouldn't we start with a single package?

recipe/meta.yaml
```diff
@@ -102,12 +103,14 @@ outputs:
...
-        - git
-        - cmake
-        - make
-        - llvm-openmp  # [osx]
+        - {{ compiler('c') }}    # [not win]
+        - {{ compiler('cxx') }}  # [not win]
+        - llvm-openmp            # [osx]
+        - m2w64-toolchain        # [win]
+        - git                    # [not win]
+        - m2-git                 # [win]
```
wagnerpeer, line 111: I am not sure if this makes a difference, but I have seen `git` as build requirement in outputs defined above, but no `m2-git`. The missing, or the additional git might cause the problems?

PrimozGodec:
@wagnerpeer thanks for your suggestion. I will try to go on with them.

> > To get the build running, we might need to reduce to a minimum working recipe. I would suggest to drop Python 3.6 and 3.9 to only try the most stable and widely supported Python versions at the moment. (see .ci_support/win_64_.yaml)
> 
> I will try to temporarily disable them
> 
> > Also, I suspect issues with applying the patches. To get more information from the command executed, we could extend the build step by defining the following two variables in the .azure-pipelines/azure-pipelines-win.yml file:
> 
> - step: |
>     # ...
>   displayName: Build recipe
>   env:
>     # ...
>     GIT_CURL_VERBOSE: 1
>     GIT_TRACE: 1

I added it. I do not see anything useful so far

> Another thing, for better readibility and traceability, I suggest to change the value of git_tag in the recipe/meta.yaml to the official tag name v1.3.0.

Done

> At last some questions:
> 
> Do we need a bld.bat to build the library?

I do not know, I copied it from a previous attempt https://github.com/conda-forge/xgboost-feedstock/pull/43. Maybe @xhochy remembers the reason for this library.

> Are we trying to build all packages (libxgboost, py-xgboost, py-xgboost-cpu, xgboost, _r-xgboost-mutex, r-xgboost and r-xgboost-cpu)?
> Shouldn't we start with a single package?

I temporarily commented out a few of them. Maybe will first only start with the first one in the next attempt. Let see what happens now.

wagnerpeer:
The error output is changing. Might be the right track 😉

PrimozGodec:
@wagnerpeer or anyone else. Does anyone know what is going on here with Windows build? This the error:
```
D:\bld\xgboost_1613478349319\work\install-libxgboost.sh: line 55: /cygdrive/c/Miniconda/Scripts/conda.exe: No such file or directory
```
What is strange here is that the error is registered for line 55 of `install-libxgboost.sh`. `install-libxgboost.sh` does not have line 55 and I also added some echoes in the script but none of them is printed. So I think something must happen before this script is executed.

wagnerpeer:
> @wagnerpeer or anyone else. Does anyone know what is going on here with Windows build? This the error:
> 
> ```
> D:\bld\xgboost_1613478349319\work\install-libxgboost.sh: line 55: /cygdrive/c/Miniconda/Scripts/conda.exe: No such file or directory
> ```
> 
> What is strange here is that the error is registered for line 55 of `install-libxgboost.sh`. `install-libxgboost.sh` does not have line 55 and I also added some echoes in the script but none of them is printed. So I think something must happen before this script is executed.

I suspect, that this might be a problem due to Windows running bash. Solution could be to initialize conda, before running the bash executable/script. Something like the following in the `.azure-pipelines/azure-pipelines-win.yml` file right before calling `conda.exe build ...`:
```bash
>> conda.exe init bash
```

PrimozGodec:
It is not helping, I found here that `bash` scripts cannot be used on windows https://docs.conda.io/projects/conda-build/en/latest/resources/build-scripts.html. Do you think it also holds for conda-forge?

wagnerpeer:
Building on Windows using shell scripts is often harder than in a "natural" Linux environment. But it should be possible.

If you look at the first link I posted in the comments, there is a discussion about conda not correctly identifying the path.
Not sure if this is the same issue, but the error log states, that it cannot find the conda.exe in the path `/cygdrive/c/Miniconda/... `

bollwyvl:
Awesome work and godspeed, would love to use this on [tpot](https://github.com/conda-forge/tpot-feedstock/issues/19).

I don't know what I don't know about getting this to build on windows, but from an `outputs` perspective, perhaps...

```yaml
outputs:
  - name: libxgboost
    script: install-libxgboost.sh  # [unix]
    script: install-libxgboost.bat  # [win]
    requirements:
      build:
        ...
        - posix  # just to be sure

```

where `install-libxgboost.bat` is a wrapper:

```bat
:: copied from https://github.com/conda-forge/go-licenses-feedstock/blob/master/recipe/bld.bat
copy "%RECIPE_DIR%\install-libxgboost.sh" .
if %errorlevel% neq 0 exit /b %errorlevel%

set PREFIX=%PREFIX:\=/%
set SRC_DIR=%SRC_DIR:\=/%
set MSYSTEM=MINGW%ARCH%
set MSYS2_PATH_TYPE=inherit
set CHERE_INVOKING=1

bash -lc "./install-libxgboost.sh"
if %errorlevel% neq 0 exit /b %errorlevel%

exit /b 0
```

Of course, go is a different beast than cmake/ninja, so YMMV.

If this worked for the first output, this approach could be made parametric off `%PKG_NAME%` (which should get updated with the output name) so that there only has to be one `install-win-wrapper.bat` checked in rather than one per output.

recipe/install-py-xgboost.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+. activate "${PREFIX}"
```
xhochy, line 3: You pribably need `activate_in_script` as we are using it in the `python-feedstock` instead: https://github.com/conda-forge/python-feedstock/blob/8fae6d37ed0e54039822db89a78ab38a76db0110/recipe/meta.yaml#L129

```suggestion
```

recipe/meta.yaml
```diff
@@ -9,30 +9,33 @@ source:
...
+  git_tag: v1.3.0
   patches:
     # xgboost patches
     - 0001-conda-Unbundle-libxgboost.-dll-dylib-so.patch
     - 0002-Fix-R-package-PKGROOT.patch
 
 build:
   number: 1
-  skip: true  # [win or linux32 or py2k]
+  skip: true  # [linux32 or py2k]
```
xhochy, line 20: You can completely remove this line now. The other two are skipped by default nowadays.
```suggestion
```

recipe/meta.yaml
```diff
@@ -41,104 +44,107 @@ outputs:
         - cmake
         - make
         - llvm-openmp  # [osx]
+        - posix
```
xhochy, line 47: `posix` should be installed by default, so this shouldn't be necessary. Otherwise, if you really need it, try to keep the build requirements here and in the top-level recipe in sync.

recipe/install-py-xgboost.sh
PrimozGodec: Would this replace the need for `install-libxgboost.bat` as it was suggested in the comment above?

xhochy: Using the generic wrapper suggested by @bollwyvl should work then, yes.

PrimozGodec:

wagnerpeer:
Great job. Thank you for all the work you have done.
I will not be able to review the result, as I have no technical expertise on xgboost.

If there is anything you need my help for, please ping me.

bollwyvl:
I'm starting to see :heavy_check_mark:s! Awesome stuff!

PrimozGodec:
@wagnerpeer thank you too and to others that helped. It is not finished yet but the progress is good. 

> I will not be able to review the result, as I have no technical expertise on xgboost.

We can ask maintainers to take a look. They should understand the package better.

recipe/meta.yaml
```diff
@@ -96,18 +101,21 @@ outputs:
...
-        - {{ compiler('c') }}
-        - {{ compiler('cxx') }}
-        - git
-        - cmake
-        - make
-        - llvm-openmp  # [osx]
+        - {{ compiler('c') }}    # [not win]
+        - {{ compiler('cxx') }}  # [not win]
+        - llvm-openmp            # [osx]
+        - m2w64-toolchain        # [win]
```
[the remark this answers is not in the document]
xhochy: The current error could also be fixed by using `clang` instead of MSYS2 but I would expect that to throw different errors. Still, this might be worth a try.
```suggestion
        - {{ compiler('r_clang') }}        # [win]
```

recipe/meta.yaml
```diff
@@ -132,13 +140,13 @@ outputs:
...
-      run:
-        - r-base
-        - {{ pin_subpackage('r-xgboost', exact=True) }}
+  # - name: r-xgboost-cpu
+  #   requirements:
+  #     host:
+  #       - r-base
+  #     run:
+  #       - r-base
+  #       - {{ pin_subpackage('r-xgboost', exact=True) }}
```
xhochy, line 149: As this is an empty package, you should be able to uncomment it.
```suggestion
  - name: r-xgboost-cpu
    requirements:
      host:
        - r-base
      run:
        - r-base
        - {{ pin_subpackage('r-xgboost', exact=True) }}
```

recipe/meta.yaml
```diff
@@ -96,18 +101,21 @@ outputs:
       string: cpu_0
 
   - name: r-xgboost
-    script: install-r-xgboost.sh
+    script: install-r-xgboost.sh  # [not win]
+    script: install-win-wrapper.bat  # [win]
```
xhochy, line 105: You could probably do a much simpler script for windows:

```
@echo on
pushd %SRC_DIR%\R-package
%R% CMD INSTALL --preclean --build .
if errorlevel 1 exit 1
```

Then you don't need a wrapper and this would then also activate the build environment correctly.

recipe/meta.yaml
isuruf: `git` is broken on windows. `m2-git` is preferred.

xhochy: Do we have an issue to track that? I'm using `git` on Windows from time to time :D

isuruf: Nope.  `git` package on windows is bundling its on msys2 packages and when conda msys2 packages are active, they mess up.

xhochy:
Seems like you can also install the R package using `cmake`: https://xgboost.readthedocs.io/en/latest/build.html#installing-the-development-version-with-visual-studio 

I'll give this a try in a separate branch.

PrimozGodec:
@xhochy thank you for testing out the option with `cmake`.

I also think there will be an issue since in the `xgboost` output this error appears. I will try to dig in this one tomorrow.
AssertionError: use-system-libxgboost is specified, but xgboost.dll is not found in: D:\bld\xgboost_1614619027633\_h_env\lib

Issue #26, opened 2019-04-30
Xgboost GPU support


RAMitchell:
Hi @aldanor, @beckermr, I am an xgboost dev mostly responsible for the GPU algorithms. We have had a few requests to get a GPU enabled package up on anaconda for linux. What would it take to make this happen? I am fairly new to anaconda, any guidance would be appreciated.

Thanks,
Rory

jakirkham:
Hi Rory, thanks for stopping by and filing this request.

Currently conda-forge does not have GPU support, but it is something we are hoping to change.

To get things started, have submitted PR ( https://github.com/conda-forge/docker-images/pull/93 ) to create a new Docker image (similar to the one used for Linux builds currently), which is based off of existing NVIDIA CUDA Docker images that include NVCC. Also have submitted PR ( https://github.com/conda-forge/staged-recipes/pull/8229 ), which creates a shim compiler package for conda-build so it can use the existing NVCC install in the Docker image easily. There are a few more things that we will want to do, but that is a good starting point.

Right now am waiting for community feedback on these two proposals. If you have any thoughts on them, please feel free to chime in.

jakirkham:
Do you have a Conda recipe you have been using for this currently?

Barring that what would be the best instructions to follow for building a GPU enabled package? [These]( https://xgboost.readthedocs.io/en/latest/build.html#building-with-gpu-support ) or something else?

Also what are your requirements in terms of GPU libraries (e.g. CUDA, NCCL, etc.) versions? Are there any other library dependencies we should be aware of?

Finally is there a good test we could run to ensure the build worked correctly?

RAMitchell:
We have a Jenkins based CI system that builds python wheels for pypy.

Here is the docker file: https://github.com/dmlc/xgboost/blob/master/tests/ci_build/Dockerfile.gpu_build

This contains appropriate nccl versions. We have been releasing builds with cuda 8.0 for maximum compatibility. We then test CPU algorithms in a minimal container (https://github.com/dmlc/xgboost/blob/master/tests/ci_build/Dockerfile.release) to ensure it still works on a system without a GPU, as well as testing the GPU algorithms on a GPU enabled container.

When I say testing, I would run the Python tests here (tests/python-gpu), disabling or enabling the multi-GPU tests as appropriate. Additionally run the google test executable testxgboost.exe, you will have to enable google tests in the cmake options to generate this.

jakirkham:
[...]

PR #84, opened 2022-02-13
add gpu support

izahn:
Checklist

Fixes https://github.com/conda-forge/xgboost-feedstock/issues/26


recipe/build.sh
```diff
@@ -3,12 +3,24 @@
...
 pushd build-target
-  cmake ${CMAKE_ARGS} \
-    -GNinja \
-    -DCMAKE_BUILD_TYPE:STRING="Release" \
-    -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON \
-    -DCMAKE_INSTALL_PREFIX:PATH="${PREFIX}" \
-    "${SRC_DIR}"
-  ninja
+if [[ ${cuda_compiler_version} != "None" ]]; then
+    cmake ${CMAKE_ARGS} \
```
xhochy, line 9: Is there any difference between these two CMake calls?

xhochy (commented):
Can you remind me how we get the `cpu` package to be the default on installation?

recipe/build.sh
izahn: The second one adds `DUSE_CUDA=ON` which is .. backwards. Ooops, will fix it!

xhochy: Can you instead use the following? 
```
if [[ ${cuda_compiler_version} != "None" ]]; then
  export CMAKE_ARGS="-DUSE_CUDA=ON ${CMAKE_ARGS}"
fi
```

That would duplicate less code.

izahn:
I think this is basically working now, but I could use some help making sure the variants are coordinated correctly. If you can take a look at the build strings and mutex packages and see if it makes sense that would be great!

recipe/meta.yaml
```diff
@@ -17,18 +17,24 @@ source:
...
     - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}                 # [cuda_compiler_version != "None"]
     - git                    # [not win]
     - m2-git                 # [win]
     - cmake
     - llvm-openmp  # [osx]
     - ninja
   host:
+    - cudatoolkit      # [cuda_compiler_version != "None"]
+    - cudnn      # [cuda_compiler_version != "None"]
```
trivialfis, line 36: I'm not familiar with conda packaging, but cudnn seems to be unnecessary?

recipe/build.sh
```diff
@@ -3,12 +3,17 @@
 set -exuo pipefail
 
 mkdir -p build-target
+
+if [[ ${cuda_compiler_version} != "None" ]]; then
+    export CMAKE_ARGS="-DUSE_CUDA=ON ${CMAKE_ARGS}"
```
trivialfis, line 8: I see that you have mentioned nccl in the patch, you might want to add `USE_NCCL=ON`, and optionally `BUILD_WITH_SHARED_NCCL=ON` to avoid the static library.

trivialfis (commented):
Excited to see progress on GPU build.  I'm not familiar with conda packaging, but feel free to ping me if there's anything I can help.

recipe/meta.yaml
izahn: Removed in https://github.com/conda-forge/xgboost-feedstock/pull/84/commits/83a0714d4a46ab8e236f62fc1efd92176833f251 thank you!

jakirkham:
Thanks Ista! 😀

Also cc-ing @hcho3

izahn:
> Can you remind me how we get the `cpu` package to be the default on installation?

From what I've been able to gather this is done by depending on `__cuda` which will cause the cpu version to be selected by default on systems without a suitable gpu. I added this in https://github.com/conda-forge/xgboost-feedstock/pull/84/commits/77429873efd1e5f3708d1ebc573ca1767119c0a3 that should do the trick.

izahn:
OK I've cleaned everything up and done my best with mutex and meta packages, and the build strings should all make sense. This is ready for review.

xhochy (approved):
Thanks @izahn !

PR #35, opened 2019-10-25
[WIP] Use one metapackage for CPU/GPU

jakirkham:
Based on discussion in issue ( https://github.com/conda-forge/xgboost-feedstock/issues/23 ), it seems that CPU/GPU selection matters mainly for package size and only really affects the C++ library. Given this information, this simplifies the recipe to have one metapackage to select between CPU/GPU builds. Note that we have **not** added GPU builds in this PR, but hopefully this should simplify the process of adding such a GPU PR in the near future.

Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

PR #132, opened 2023-09-14
Release 2.0.3

PrimozGodec:
Checklist


PrimozGodec:

aldanor:
Lots of failure here... cuda builds are failing like
```
$SRC_DIR/src/common/hist_util.cu(145): error: no instance of function template "thrust::make_zip_iterator" matches the argument list
            argument types are: (xgboost::Entry *, float *)

$SRC_DIR/src/common/hist_util.cu(146): error: no instance of function template "thrust::make_zip_iterator" matches the argument list
            argument types are: (xgboost::Entry *, float *)

$SRC_DIR/src/common/hist_util.cu(147): error: no instance of overloaded function "dh::SegmentedUnique" matches the argument list
            argument types are: (size_t *, size_t *, <error-type>, <error-type>, size_t *, <error-type>, lambda [](const Pair &, const Pair &)->__nv_bool)
```

Also OSX builds are failing
```
In file included from $SRC_DIR/src/c_api/c_api.cc:28:
$SRC_DIR/src/c_api/../data/proxy_dmatrix.h:137:25: error: 'any_cast<std::shared_ptr<xgboost::data::CSRArrayAdapter>>' is unavailable: introduced in macOS 10.13 - see https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
      auto value = std::any_cast<std::shared_ptr<CSRArrayAdapter>>(proxy->Adapter());
                        ^
$BUILD_PREFIX/bin/../include/c++/v1/any:638:12: note: 'any_cast<std::shared_ptr<xgboost::data::CSRArrayAdapter>>' has been explicitly marked unavailable here
_ValueType any_cast(any && __v)
           ^
In file included from $SRC_DIR/src/c_api/c_api.cc:28:
$SRC_DIR/src/c_api/../data/proxy_dmatrix.h:145:25: error: 'any_cast<std::shared_ptr<xgboost::data::ArrayAdapter>>' is unavailable: introduced in macOS 10.13 - see https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
      auto value = std::any_cast<std::shared_ptr<ArrayAdapter>>(proxy->Adapter())->Value();
                        ^
$BUILD_PREFIX/bin/../include/c++/v1/any:638:12: note: 'any_cast<std::shared_ptr<xgboost::data::ArrayAdapter>>' has been explicitly marked unavailable here
_ValueType any_cast(any && __v)
           ^
In file included from $SRC_DIR/src/c_api/c_api.cc:28:
$SRC_DIR/src/c_api/../data/proxy_dmatrix.h:148:25: error: 'any_cast<std::shared_ptr<xgboost::data::ArrayAdapter>>' is unavailable: introduced in macOS 10.13 - see https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
      auto value = std::any_cast<std::shared_ptr<ArrayAdapter>>(proxy->Adapter());
                        ^
$BUILD_PREFIX/bin/../include/c++/v1/any:638:12: note: 'any_cast<std::shared_ptr<xgboost::data::ArrayAdapter>>' has been explicitly marked unavailable here
_ValueType any_cast(any && __v)
           ^
```

elgalu:
any updates?

hcho3:
Let me try to take a look when I get back from vacation (next Tuesday)

jakirkham:
@conda-forge-admin , please re-render

jakirkham:
@conda-forge-admin , please re-render

jakirkham:
@conda-forge-admin , please re-render

jakirkham:
@conda-forge-admin , please re-render

jakirkham:
@conda-forge-admin , please re-render

jakirkham:
How should we handle [these optional dependencies]( https://github.com/dmlc/xgboost/blob/ea9f09716bd426f235ce785d6286e0ad26350ce1/python-package/pyproject.toml#L39-L45 )?

```TOML
[project.optional-dependencies]
pandas = ["pandas"]
scikit-learn = ["scikit-learn"]
dask = ["dask", "pandas", "distributed"]
datatable = ["datatable"]
plotting = ["graphviz", "matplotlib"]
pyspark = ["pyspark", "scikit-learn", "cloudpickle"]
```

Currently we just have `scikit-learn` from this list. Should we keep that or drop it (and let others install it)? What about the other ones in this list?

jakirkham:
Not sure what version of Thrust is needed here, but it looks like the version that ships with CTK 11.2 is too old. Does anyone here know what version is needed?

We can add Thrust as its own package or use CCCL (where Thrust has been moved into). Either way we can get a newer version, which should fix this issue.

jakirkham:
Also seeing this error on Windows. Maybe there is more needed when specifying where the library is

```python
   Traceback (most recent call last):
     File "D:\bld\xgboost-split_1697601995583\_h_env\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
       main()
     File "D:\bld\xgboost-split_1697601995583\_h_env\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
       json_out['return_val'] = hook(**hook_input['kwargs'])
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     File "D:\bld\xgboost-split_1697601995583\_h_env\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 152, in prepare_metadata_for_build_wheel
       whl_basename = backend.build_wheel(metadata_directory, config_settings)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     File "D:\bld\xgboost-split_1697601995583\work\python-package\packager\pep517.py", line 79, in build_wheel
       libxgboost = locate_or_build_libxgboost(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
     File "D:\bld\xgboost-split_1697601995583\work\python-package\packager\nativelib.py", line 138, in locate_or_build_libxgboost
       raise RuntimeError(
   RuntimeError: use_system_libxgboost was specified but xgboost.dll is not found in D:\bld\xgboost-split_1697601995583\_h_env\lib
```

hcho3:
We can probably remove CUDA 11.2 and build for CUDA 11.8 and 12.0 only.

As for locating the library, the following lines need to be patched to use `sys_base_prefix / "Library" / "lib"` instead:
```python
sys_base_prefix = pathlib.Path(sys.base_prefix).absolute().resolve()
libxgboost_sys = sys_base_prefix / "lib" / _lib_name()
```
(https://github.com/dmlc/xgboost/blob/f7da938458d4cb012e08fec6ae60a4ec59543b1b/python-package/packager/nativelib.py#L135-L136)

Here's a useful reference: https://github.com/dmlc/treelite/blob/6cd029570b762ffe04acf1019e9599de3b75339c/python/packager/nativelib.py#L140-L144

I can prepare a patch tomorrow.

hcho3:
@jakirkham Can you add this patch? I am going to submit this patch to the upcoming XGBoost patch release 2.0.1.

```patch
diff --git a/python-package/packager/nativelib.py b/python-package/packager/nativelib.py
index ff38fa11..164bd09e 100644
--- a/python-package/packager/nativelib.py
+++ b/python-package/packager/nativelib.py
@@ -132,16 +132,28 @@ def locate_or_build_libxgboost(

     if build_config.use_system_libxgboost:
         # Find libxgboost from system prefix
-        sys_base_prefix = pathlib.Path(sys.base_prefix).absolute().resolve()
-        libxgboost_sys = sys_base_prefix / "lib" / _lib_name()
-        if not libxgboost_sys.exists():
-            raise RuntimeError(
-                f"use_system_libxgboost was specified but {_lib_name()} is "
-                f"not found in {libxgboost_sys.parent}"
-            )
-
-        logger.info("Using system XGBoost: %s", str(libxgboost_sys))
-        return libxgboost_sys
+        sys_prefix = pathlib.Path(sys.prefix)
+        sys_prefix_candidates = [
+            sys_prefix / "lib",
+            # Paths possibly used on Windows
+            sys_prefix / "bin",
+            sys_prefix / "Library",
+            sys_prefix / "Library" / "bin",
+            sys_prefix / "Library" / "lib",
+        ]
+        sys_prefix_candidates = [
+            p.expanduser().resolve() for p in sys_prefix_candidates
+        ]
+        for candidate_dir in sys_prefix_candidates:
+            libtreelite_sys = candidate_dir / _lib_name()
+            if libtreelite_sys.exists():
+                logger.info("Using system XGBoost: %s", str(libtreelite_sys))
+                return libtreelite_sys
+        raise RuntimeError(
+            f"use_system_libxgboost was specified but {_lib_name()} is "
+            f"not found. Paths searched (in order): \n"
+            + "\n".join([f"* {str(p)}" for p in sys_prefix_candidates])
+        )

     libxgboost = locate_local_libxgboost(toplevel_dir, logger=logger)
     if libxgboost is not None:
diff --git a/python-package/xgboost/libpath.py b/python-package/xgboost/libpath.py
index 0437f3a4..36a92cbc 100644
--- a/python-package/xgboost/libpath.py
+++ b/python-package/xgboost/libpath.py
@@ -31,16 +31,15 @@ def find_lib_path() -> List[str]:
     ]

     if sys.platform == "win32":
-        if platform.architecture()[0] == "64bit":
-            dll_path.append(os.path.join(curr_path, "../../windows/x64/Release/"))
-            # hack for pip installation when copy all parent source
-            # directory here
-            dll_path.append(os.path.join(curr_path, "./windows/x64/Release/"))
-        else:
-            dll_path.append(os.path.join(curr_path, "../../windows/Release/"))
-            # hack for pip installation when copy all parent source
-            # directory here
-            dll_path.append(os.path.join(curr_path, "./windows/Release/"))
+        # On Windows, Conda may install libs in different paths
+        dll_path.extend(
+            [
+                os.path.join(sys.prefix, "bin"),
+                os.path.join(sys.prefix, "Library"),
+                os.path.join(sys.prefix, "Library", "bin"),
+                os.path.join(sys.prefix, "Library", "lib"),
+            ]
+        )
         dll_path = [os.path.join(p, "xgboost.dll") for p in dll_path]
     elif sys.platform.startswith(("linux", "freebsd", "emscripten")):
         dll_path = [os.path.join(p, "libxgboost.so") for p in dll_path]
```

elgalu:
yes team please stop building for CUDA 11.2 that's from 2020. CUDA >= 11.8 should be aligned with the current status quo.

xhochy:
We should drop all of them and provide the extras as additional outputs.

jakirkham:
It sounds like there is some consensus around leaving optional dependency out. As this is 2.0.0, think it is reasonable to drop `scikit-learn` at the same time

While I can see why it would be useful to have additional outputs with optional dependencies, this is also more maintenance effort that might be spent better in other ways. Also it doesn't move the needle a whole lot for users, who could just install these themselves

Plus we haven't heard a lot of interest from users up to this point to handle the optional dependencies. So am guessing they are ok with the status quo

recipe/0002-libpath.patch
```diff
@@ -0,0 +1,86 @@
+From 2cdc7f71d586cbb15322b81c17b98cf99cc98bbe Mon Sep 17 00:00:00 2001
```
hcho3, line 1: This patch will be part of the upcoming patch release 2.0.1 of XGBoost.

jakirkham:
Since this is 2.0.0, this seems like a good opportunity to cutdown on some complexity in this recipe

In particular, we have a bunch of metapackages to select between CPU & GPU packages. However AFAICT only `libxgboost` has CPU & GPU bits. Would it be possible to only have one metapackage that affects `libxgboost` directly? Perhaps borrowing from past work ( https://github.com/conda-forge/xgboost-feedstock/pull/35 )

Related to that. Does the ABI surface of `libxgboost` change between CPU & GPU builds? If not can, we reduce the number of Python & R package builds accordingly to be reused for both CPU & GPU

Another consideration is simplifying the Python (and maybe R?) package builds. IIUC the Python package is pure Python. So could use `noarch: python` like so ( https://github.com/conda-forge/xgboost-feedstock/pull/128 ). Is the R code also pure R code (without C/C++)? If so, maybe we can move these to `noarch` to cutdown on build time

This complexity affects things like the ability of the bot to perform updates and migrations reliably. Here's a recent example: https://github.com/regro/cf-scripts/issues/1890

Anyways the build time is kind of long, the recipe has grown complex to handle the different interests and concerns here. Think there is some value in reigning in the complexity to cutdown on build time, improve the maintenance experience, and make the packages a bit more user friendly

hcho3:
@jakirkham I agree with the need for simplifying the recipe.

recipe/meta.yaml
```diff
@@ -56,6 +57,7 @@ outputs:
       host:
         - nccl         # [linux and cuda_compiler != "None"]
       run:
+        - __osx >={{ MACOSX_DEPLOYMENT_TARGET|default("10.15") }}  # [osx and x86_64]
         - __cuda                   # [cuda_compiler != "None"]
         {% if (cuda_compiler_version or "").startswith("11") %}
         - cuda-version >={{ cuda_compiler_version }},<12
```
jakirkham, line 63: Think we may also want to update this to allow CUDA 11.2 (if we are reusing CUDA 11.8 builds on older CUDA versions)

```suggestion
        - cuda-version >=11.2,<12
```

hcho3: Should we allow CUDA 11.2 at all?

jakirkham: At present I would like to make sure we are supporting users of CUDA 11.2+ somehow. If reusing CUDA 11.8 builds on 11.2+ does that, then would help simplify things here

hcho3:
@jakirkham Can I get a review? The CI is now all green.

PrimozGodec:
@hcho3, thank you for the fixes. It seems that it requires minor conflict-solving because of changes at the master.

I would be very happy if someone could review that soon to make version 2 available on conda-forge.

jayqi:
Hi @jakirkham, @hcho3, @PrimozGodec. It's been a few weeks since the last activity, and I was hoping to check if any of you might have an update on when this release might be completed.

hcho3:
Can I get another round of review?

jakirkham:
Thanks Philip! 🙏

Yes can look tomorrow

recipe/meta.yaml
```diff
@@ -90,6 +101,7 @@ outputs:
         - numpy
         - scipy
         - scikit-learn
+        - __osx >={{ MACOSX_DEPLOYMENT_TARGET|default("10.15") }}  # [osx and x86_64]
```
xhochy, line 104: This line should only be on the `libxgboost` output.

recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+MACOSX_DEPLOYMENT_TARGET:  # [osx and x86_64]
```
xhochy, line 1: What's the reason for increasing the deployment target? Can you please add a comment on why passing `_LIBCPP_DISABLE_AVAILABILITY` wasn't sufficient?

recipe/build.sh
```diff
@@ -4,8 +4,16 @@ set -exuo pipefail
 
 mkdir -p build-target
 
+CMAKE_ARGS=()
```
xhochy, line 7: Please don't unset the `CMAKE_ARGS` that come in from the compiler activation.

xhochy (changes requested):
A lot of flags seems to be duplicated here, they should only be in one place.

recipe/conda_build_config.yaml
hcho3: XGBoost uses C++17 features such as `std::filesystem`. Is passing `_LIBCPP_DISABLE_AVAILABILITY` sufficient to enable C++17 features?

hcho3: @jakirkham Any inputs on this?

xhochy: Yes, that should be sufficient.

hcho3: Let me try removing `MACOSX_DEPLOYMENT_TARGET` flags and see if the build still succeeds.

hcho3:
@conda-forge-admin , please re-render

hcho3:
@xhochy I addressed all your comments. Can you look and see if the changes look reasonable?

PR #167, opened 2024-06-29
Upgrade to XGBoost 2.1.0

hcho3:
Checklist

Fixes https://github.com/conda-forge/xgboost-feedstock/issues/172


hcho3:

jakirkham:
@conda-forge-admin , please restart CI

jakirkham:
Did the source layout change somehow? CI is noting that it had trouble finding the `xgboost` directory

Not seeing this issue when building for the previous release

trivialfis:
The `rabit` directory is removed.

recipe/install-libxgboost.sh
```diff
@@ -30,5 +30,4 @@ mkdir -p ${LIBDIR} ${INCDIR}/xgboost ${BINDIR} || true
 cp ${SRC_DIR}/xgboost${EXEEXT} ${BINDIR}/
 cp ${SRC_DIR}/lib/${XGBOOSTDSO} ${SODIR}/
 cp -Rf ${SRC_DIR}/include/xgboost ${INCDIR}/
-cp -Rf ${SRC_DIR}/rabit/include/rabit ${INCDIR}/xgboost/
```
jakirkham, line 33: So should we drop this line?

hcho3: Yes

jakirkham:
Still seeing the same error [on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=965448&view=logs&j=cacb33e2-4f35-58d8-c6f0-bc569fa20097&t=5d96f9ff-2d0d-52ea-cc90-ffadc2f5efed&l=1149 ):

```
cp: cannot stat ‘/home/conda/feedstock_root/build_artifacts/xgboost-split_1719641630231/work/xgboost’: No such file or directory
```

Are there other layout changes that occurred?

recipe/meta.yaml
```diff
@@ -1,6 +1,6 @@
 {% set name = "xgboost" %}
-{% set version = "2.0.3" %}
-{% set build_number = 4 %}
+{% set version = "2.1.0" %}
+{% set build_number = 1 %}
```
jakirkham, line 3:
```suggestion
{% set build_number = 0 %}
```

recipe/install-libxgboost.sh
jakirkham: Am still seeing [`rabit` in 2.1.0]( https://github.com/dmlc/xgboost/tree/v2.1.0/rabit ). Is it not suppose to be there or does it just get removed later?

recipe/install-libxgboost.sh
```diff
@@ -26,9 +26,9 @@ else
   EXEEXT=
 fi
 
+ls ${SRC_DIR}
 mkdir -p ${LIBDIR} ${INCDIR}/xgboost ${BINDIR} || true
 cp ${SRC_DIR}/xgboost${EXEEXT} ${BINDIR}/
```
jakirkham, line 31: Is this executable still around or did it get removed in 2.1.0?

hcho3: Yes, let's remove it

jakirkham: Cool that should fix CI: https://github.com/conda-forge/xgboost-feedstock/pull/167#issuecomment-2198017977

recipe/install-libxgboost.sh
jakirkham: Looks like it is just used in a couple of C++ source files:

```
% git grep 'rabit/internal/socket\.h'
src/collective/loop.cc:#include "rabit/internal/socket.h"      // for PollHelper
src/collective/socket.cc:#include "rabit/internal/socket.h"      // for PollHelper
src/collective/tracker.cc:#include "rabit/internal/socket.h"
```

Plus it is removed in `master`

So seems ok to drop even though there is still something left in there. Users shouldn't need it

recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,2 @@
+r_base:
+  - 4.3
```
jakirkham, line 2: Would change this to a `skip` in `meta.yaml`

```yaml
  skip: true  # [r_base != "4.3"]
```

trivialfis:
Is It possible that we skip R build on this release? The rewrite is still on going.

hcho3:
> Is It possible that we skip R build on this release? The rewrite is still on going.

Not sure if commenting out the R recipe is the right way to approach this.

jakirkham:
@conda-forge-admin , please re-render

hcho3:

jakirkham:
@conda-forge-admin , please re-render

recipe/meta.yaml
```diff
@@ -24,6 +24,7 @@ source:
 build:
   number: {{ build_number }}
   skip: True  # [cuda_compiler_version == "11.2"]
+  skip: True  # [r_base != "4.3"]
```
jakirkham, line 27: Would it be possible to retry this?

```suggestion
```

hcho3: Without this line, Windows will fail because it doesn't yet have R 4.3.

jakirkham: Ok so XGBoost needs at least R 4.3?

If so, could we rewrite this to just skip [R 4.1 used on Windows]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/a724f060a0629f6be7b16c06cdce06467788dee5/recipe/conda_build_config.yaml#L831-L833 )? That way it will work with R 4.4 when that migration comes around

```suggestion
  skip: True  # [r_base == "4.1"]
```

Also would suggest we move this `skip` to the `r-xgboost*` packages specifically. That way we don't skip Windows altogether and just skip R on Windows

jameslamb:
@jakirkham pointed out that the failing `aarch64` build ([build link](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=978588&view=logs&jobId=d1321064-f6c3-56b7-0172-6d994d01c836&j=e9afaa34-1a0f-534e-78fc-fae528ccd915&t=48395e60-bc5f-5e45-7455-d404f0c95f3a)) has this:

```text
Reloading output folder (local): ...working... done
Solving environment (_test_env): ...working... failed
WARNING: failed to get package records, retrying.  exception was: Unsatisfiable dependencies for platform linux-64: {MatchSpec("libxgboost[version='>=2.1.0,<2.1.1.0a0']"), MatchSpec("py-xgboost==2.1.0=cuda120_pyhee1328b_0"), MatchSpec(">=2.1.0,<2.1.1.0a0")}
Encountered problems while solving:
  - nothing provides requested libxgboost >=2.1.0,<2.1.1.0a0
  - nothing provides libxgboost >=2.1.0,<2.1.1.0a0 needed by py-xgboost-2.1.0-cuda120_pyhee1328b_0

Could not solve for environment specs
The following packages are incompatible
├─ libxgboost >=2.1.0,<2.1.1.0a0  does not exist (perhaps a typo or a missing channel);
└─ py-xgboost 2.1.0.* cuda120_pyhee1328b_0 is not installable because it requires
   └─ libxgboost >=2.1.0,<2.1.1.0a0 , which does not exist (perhaps a missing channel).
```

Earlier in that build, I do see the right thing happening.

```text
Packaging py-xgboost
...

The following NEW packages will be INSTALLED:
...
   libxgboost:            2.1.0-cuda120_haef42e6_0   local      
...
```

So I'm not sure what's happening, but just wanted to help narrow it down. This error is coming from the point where conda-build is building the `xgboost` package specifically.

Around the failing test logs, I noticed this:

```text
TEST START: /home/conda/feedstock_root/build_artifacts/linux-aarch64/_py-xgboost-mutex-2.0-gpu_0.conda
WARNING: Multiple meta files found. The meta.yaml file in the base directory (/tmp/tmp8n2pxzql/info/recipe) will be used.
Nothing to test for: /home/conda/feedstock_root/build_artifacts/linux-aarch64/_py-xgboost-mutex-2.0-gpu_0.conda
TEST START: /home/conda/feedstock_root/build_artifacts/linux-aarch64/_r-xgboost-mutex-2.0-gpu_0.conda
```

Notice the version numbers there are `2.0`, but this PR is for `2.1.0`. I don't really understand what those `*-mutex` packages do, but just noting in case it helps... should their version be bumped to `2.1.0`?

https://github.com/conda-forge/xgboost-feedstock/blob/8ff49727db706331eb68cccf40dff8b14c1ec86a/recipe/meta.yaml#L74

https://github.com/conda-forge/xgboost-feedstock/blob/8ff49727db706331eb68cccf40dff8b14c1ec86a/recipe/meta.yaml#L153

jakirkham:
In looking at James' observation above noticed that we were not consistently adding `pin_subpackage` to `host` and `run`. Am wondering if this confused the build tooling when deciding on the build order. Tried to add `pin_subpackage` to both `host` and `run` (if it was used in `run` only before). Let's see if that helps

jakirkham:
In Linux `ppc64le` builds, we are seeing [this error on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=980024&view=logs&j=9d6b5b0e-8566-52e2-1634-02118cea4eac&t=2c35c673-9b44-5873-66a2-1f997c4d4ca8&l=1198 ):

```python
Warning: rpath /home/conda/feedstock_root/build_artifacts/xgboost-split_1721148451617/_build_env/lib is outside prefix /home/conda/feedstock_root/build_artifacts/xgboost-split_1721148451617/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho (removing it)
   INFO: sysroot: '/home/conda/feedstock_root/build_artifacts/xgboost-split_1721148451617/_build_env/x86_64-conda-linux-gnu/sysroot/' files: '['usr/share/zoneinfo', 'usr/share/locale/zh_TW/LC_MESSAGES/libc.mo', 'usr/share/locale/zh_CN/LC_MESSAGES/libc.mo', 'usr/share/locale/vi/LC_MESSAGES/libc.mo']'
   INFO: sysroot: '/home/conda/feedstock_root/build_artifacts/xgboost-split_1721148451617/_build_env/powerpc64le-conda-linux-gnu/sysroot/' files: '['usr/share/zoneinfo', 'usr/share/locale/zh_TW/LC_MESSAGES/libc.mo', 'usr/share/locale/zh_CN/LC_MESSAGES/libc.mo', 'usr/share/locale/vi/LC_MESSAGES/libc.mo']'
   INFO (libxgboost,lib/libxgboost.so): Needed DSO x86_64-conda-linux-gnu/sysroot/lib64/libpthread.so.0 found in CDT/compiler package conda-forge/noarch::sysroot_linux-64==2.17=h4a8ded7_16
   INFO (libxgboost,lib/libxgboost.so): Needed DSO lib/libgomp.so.1 found in conda-forge/linux-ppc64le::_openmp_mutex==4.5=2_gnu
   INFO (libxgboost,lib/libxgboost.so): Needed DSO lib/libstdc++.so.6 found in conda-forge/linux-ppc64le::libstdcxx-ng==14.1.0=h262982c_0
   INFO (libxgboost,lib/libxgboost.so): Needed DSO x86_64-conda-linux-gnu/sysroot/lib64/libm.so.6 found in CDT/compiler package conda-forge/noarch::sysroot_linux-64==2.17=h4a8ded7_16
   INFO (libxgboost,lib/libxgboost.so): Needed DSO lib/libgcc_s.so.1 found in conda-forge/linux-ppc64le::libgcc-ng==14.1.0=h31e42bb_0
   INFO (libxgboost,lib/libxgboost.so): Needed DSO x86_64-conda-linux-gnu/sysroot/lib64/libc.so.6 found in CDT/compiler package conda-forge/noarch::sysroot_linux-64==2.17=h4a8ded7_16
  ERROR (libxgboost,lib/libxgboost.so): $RPATH/ld64.so.2 not found in packages, sysroot(s) nor the missing_dso_whitelist.
.. is this binary repackaging?
```

Appears to be upstream issue: https://github.com/conda/conda-build/issues/5403

This will be fixed in the next `conda-build` release. We can workaround it for now

jakirkham:
@conda-forge-admin, please re-render

jakirkham:
@conda-forge-admin, please re-render

jakirkham:
@conda-forge-admin, please re-render

hcho3:
@conda-forge-admin, please re-render

hcho3:
The selector expression for `r_base` doesn't seem to work for the Windows platform. See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=980389&view=logs&jobId=7764172a-9256-5214-5ef5-618c061b9299&j=7e77ae57-7522-554c-2968-3fa3255b91c4&t=79b292e0-4719-5269-0bd3-327eef2bb687, where R 3.4 is chosen.

For now, I am explicitly skipping Windows for the R package.

hcho3:
@conda-forge-admin, please re-render

jakirkham:
@conda-forge-admin, please re-render

jakirkham:
@conda-forge-admin, please re-render

recipe/bld.bat
```diff
@@ -1,8 +1,14 @@
...
 
-if NOT "%cuda_compiler_version%"=="None" (
+if not "%cuda_compiler_version%" == "None" (
     set "CMAKE_ARGS=-DUSE_CUDA=ON %CMAKE_ARGS%"
+
+    :: Workaround CUB build errors on CUDA 11.8
+    if not "%cuda_compiler_version%" == "11.8" (
+        set "CXXFLAGS=%CXXFLAGS% -DWIN32_LEAN_AND_MEAN"
+        set "NVCC_APPEND_FLAGS=%NVCC_APPEND_FLAGS% -DWIN32_LEAN_AND_MEAN"
+    )
```
jakirkham, line 11: Think we need something like this for Windows CUDA 11.8. That or we use a newer version of CUB than the CTK includes

xref: https://github.com/facebookresearch/pytorch3d/issues/1614#issuecomment-1719514892

jakirkham:
@conda-forge-admin, please re-render

recipe/bld.bat
```diff
@@ -1,16 +1,23 @@
...
+mkdir %SRC_DIR%\build
+pushd %SRC_DIR%\build
+
+if not "%cuda_compiler_version%" == "None" (
     set "CMAKE_ARGS=-DUSE_CUDA=ON %CMAKE_ARGS%"
+
+    :: Workaround CUB build errors on CUDA 11.8
+    if "%cuda_compiler_version%" == "11.8" (
+        set "CUDAFLAGS=%CUDAFLAGS% -DWIN32_LEAN_AND_MEAN"
+    )
```
jakirkham, line 12: So this helps the Windows CUDA 11.8 build progress

Though now see a new error

```
%SRC_DIR%\src\data\adapter.h(289): here



Internal error: assertion failed at: "overload.c", line 28946 in deduce_class_template_args





1 catastrophic error detected in the compilation of "D:/bld/xgboost-split_1721471361683/work/src/data/ellpack_page_source.cu".

Compilation aborted.

ellpack_page_source.cu

nvcc error   : 'cudafe++' died with status 0xC0000409 
```

jakirkham:
@conda-forge-admin, please re-render

recipe/meta.yaml
```diff
@@ -63,17 +68,25 @@ outputs:
         - llvm-openmp  # [osx]
         - libgomp      # [linux]
       host:
+        - cccl         # [cuda_compiler != "None"]
```
jakirkham, line 71: Added CCCL. However Windows CUDA 11.8 is still using CUB, etc. from CTK. Does XGBoost have an option to specify where to get CUB, etc. from? Would like to use CCCL instead to see if that fixes the issue

hcho3: There used to be a CMake option `BUILD_WITH_CUDA_CUB` but it was removed in 2.0.0 (https://github.com/dmlc/xgboost/pull/8888).

hcho3: I see two alternatives:
1. Require CUDA 12.0 for Windows target OR
2. Add a patch to restore `BUILD_WITH_CUDA_CUB` option in XGBoost.

trivialfis: `BUILD_WITH_CUDA_CUB` was there when XGBoost had cub as its submodule. We can't restore it since CCCL is a third-party package. We will have to look into how to ask CMake to prioritize the cub from CCCL over cub in CTK.

bdice: `find_package(CCCL)` should find the conda package, in my recollection.

hcho3: About to send a patch to apply https://github.com/dmlc/xgboost/pull/10624

jakirkham:
@conda-forge-admin , please re-render

jakirkham:
Thanks Hyunsu! 🙏

Sorry if I goofed applying that patch, was just wanting to see how things behaved here with your improvements 🙂

hcho3:
It seems like the Windows build is still picking up CUB from CTK, not CCCL?
```
2024-07-23T00:25:20.2961508Z C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include\cub/device/dispatch/dispatch_segmented_sort.cuh(338): error: invalid combination of type specifiers
2024-07-23T00:25:20.2962328Z 
2024-07-23T00:25:20.2962959Z C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include\cub/device/dispatch/dispatch_segmented_sort.cuh(338): error: expected an identifier
2024-07-23T00:25:20.2963502Z 
2024-07-23T00:25:20.2963870Z 
2024-07-23T00:25:20.2964241Z 
2024-07-23T00:25:20.2964873Z C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include\cub/device/dispatch/dispatch_segmented_sort.cuh(379): error: expected a member name
2024-07-23T00:25:20.2965380Z 
2024-07-23T00:25:20.2965776Z 
2024-07-23T00:25:20.2966285Z 
2024-07-23T00:25:20.2967073Z 3 errors detected in the compilation of "D:/bld/xgboost-split_1721693173010/work/src/common/quantile.cu".
```

jakirkham:
Right we still need the updates from your patch for the reasons you noted ( https://github.com/dmlc/xgboost/pull/10624#issuecomment-2243993172 )

Do you want to push those changes here?

jakirkham:
Thanks Hyunsu! 🙏

It does look like CMake can [find CCCL (and associated components)]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=985622&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=0c979b57-e97b-5e0a-363d-f964823a2ab0&l=1174 )

```
-- Found CUDAToolkit: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include (found version "11.8.89")
-- Found libcudacxx: D:/bld/xgboost-split_1721698119331/_h_env/Library/lib/cmake/libcudacxx/libcudacxx-config.cmake (found suitable version "2.5.0.0", minimum required is "2.5.0.0")
-- Found Thrust: D:/bld/xgboost-split_1721698119331/_h_env/Library/lib/cmake/thrust/thrust-config.cmake (found suitable exact version "2.5.0.0")
-- Found CUB: D:/bld/xgboost-split_1721698119331/_h_env/Library/lib/cmake/cub/cub-config.cmake (found suitable version "2.5.0.0", minimum required is "2.5.0.0")
-- Found CCCL: D:/bld/xgboost-split_1721698119331/_h_env/Library/lib/cmake/cccl/cccl-config.cmake (found version "2.5.0.0")
```

However we are still seeing an [error showing CUDA 11.8 was used]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=985622&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=0c979b57-e97b-5e0a-363d-f964823a2ab0&l=4847 )

```
C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include\cub/device/dispatch/dispatch_segmented_sort.cuh(338): error: invalid combination of type specifiers

C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include\cub/device/dispatch/dispatch_segmented_sort.cuh(338): error: expected an identifier



C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/include\cub/device/dispatch/dispatch_segmented_sort.cuh(379): error: expected a member name



3 errors detected in the compilation of "D:/bld/xgboost-split_1721698119331/work/src/common/quantile.cu".
```

Are there any other places where we need to specify the CCCL headers should be used?

jakirkham:
Some other things we can try:

1. Use verbose builds to see what flags are used compilation steps
2. Output any additional error info from where the builds fail
3. Add `thrust_create_target` to the patch ( https://github.com/dmlc/xgboost/pull/10624#discussion_r1687320295 )
4. Make sure the header search path includes CCCL before the CTK
5. Verify the CCCL header path is being passed to NVCC

hcho3:
Log from the failing Windows build:

2024-07-23T19:46:02.7819177Z [110/151] C:\PROGRA~1\NVIDIA~2\CUDA\v11.8\bin\nvcc.exe -forward-unknown-to-host-compiler -ccbin=C:\PROGRA~1\MICROS~2\2022\ENTERP~1\VC\Tools\MSVC\1429~1.301\bin\HostX64\x64\cl.exe -DDMLC_CORE_USE_CMAKE -DDMLC_LOG_CUSTOMIZE=1 -DDMLC_USE_CXX11=1 -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -DXGBOOST_MM_PREFETCH_PRESENT=1 -DXGBOOST_USE_CUDA=1 -D__USE_XOPEN2K8 -I%SRC_DIR%\include -I%SRC_DIR%\dmlc-core\include -I%SRC_DIR%\rabit\include -I%SRC_DIR%\gputreeshap -I"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.8\include" -I%SRC_DIR%\build\dmlc-core\include -I%PREFIX%\Library\include -D_WINDOWS -Xcompiler=" /GR /EHsc" -Xcompiler="-O2 -Ob2" -DNDEBUG -std=c++17 "--generate-code=arch=compute_50,code=[sm_50]" "--generate-code=arch=compute_60,code=[sm_60]" "--generate-code=arch=compute_70,code=[sm_70]" "--generate-code=arch=compute_80,code=[sm_80]" "--generate-code=arch=compute_90,code=[sm_90]" "--generate-code=arch=compute_90,code=[compute_90]" -Xcompiler=-MT -D_CRT_SECURE_NO_WARNINGS -D_CRT_SECURE_NO_DEPRECATE --expt-extended-lambda --expt-relaxed-constexpr -Xcompiler=-openmp -Xfatbin=-compress-all --default-stream per-thread -lineinfo -Xcompiler=/utf-8 -MD -MT src\CMakeFiles\objxgboost.dir\common\quantile.cu.obj -MF src\CMakeFiles\objxgboost.dir\common\quantile.cu.obj.d -x cu -c %SRC_DIR%\src\common\quantile.cu -o src\CMakeFiles\objxgboost.dir\common\quantile.cu.obj -Xcompiler=-Fdsrc\CMakeFiles\objxgboost.dir\,-FS
2024-07-23T19:46:02.7825791Z FAILED: src/CMakeFiles/objxgboost.dir/common/quantile.cu.obj

jakirkham:
Thanks Hyunsu!

Will highlight a segment relevant to us

```batch
-I"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v11.8\include" ... -I%PREFIX%\Library\include 
```

It looks like the `host` environment includes are added after the CTK's includes. Can we flip that order?

recipe/meta.yaml
```diff
@@ -22,7 +22,6 @@ source:
 
 build:
   number: {{ build_number }}
-  noarch: python
```
jakirkham: This was here to make sure we don't rebuild for every Python version as the packages using Python are `noarch`

The lint here appears to be mistaken: https://github.com/conda-forge/conda-smithy/issues/1887#issuecomment-2244323850

.ci_support/linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11.yaml
```diff
@@ -32,7 +32,11 @@ pin_run_as_build:
     min_pin: x.x
     max_pin: x.x
 python:
+- 3.10.* *_cpython
+- 3.11.* *_cpython
 - 3.12.* *_cpython
+- 3.8.* *_cpython
+- 3.9.* *_cpython
```
jakirkham, line 39: Note that we now get a build matrix over `python` without the top-level `noarch`

Meaning that we rebuild each package using `python` 5 times (even though each package is identical)

jakirkham (commented):
Added some more context on the top-level `noarch: python` below

Hopefully we can find a better way to resolve the lint

jakirkham:
Looks like we are getting a compile error [on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=986451&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=841356e0-85bb-57d8-dbbc-852e683d1642&l=926 ):

```c
[9/150] Building CXX object src/CMakeFiles/objxgboost.dir/c_api/c_api.cc.o
FAILED: src/CMakeFiles/objxgboost.dir/c_api/c_api.cc.o 
$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-c++ -DDMLC_CORE_USE_CMAKE -DDMLC_LOG_CUSTOMIZE=1 -DDMLC_USE_CXX11=1 -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -DXGBOOST_BUILTIN_PREFETCH_PRESENT=1 -DXGBOOST_MM_PREFETCH_PRESENT=1 -DXGBOOST_USE_CUDA=1 -DXGBOOST_USE_NCCL=1 -D_MWAITXINTRIN_H_INCLUDED -D__USE_XOPEN2K8 -I$SRC_DIR/include -I$SRC_DIR/dmlc-core/include -I$SRC_DIR/rabit/include -I$SRC_DIR/gputreeshap -I$PREFIX/include -I$SRC_DIR/build-target/dmlc-core/include -D_LIBCPP_DISABLE_AVAILABILITY -O3 -DNDEBUG -std=gnu++17 -fPIC -funroll-loops -fopenmp -MD -MT src/CMakeFiles/objxgboost.dir/c_api/c_api.cc.o -MF src/CMakeFiles/objxgboost.dir/c_api/c_api.cc.o.d -o src/CMakeFiles/objxgboost.dir/c_api/c_api.cc.o -c $SRC_DIR/src/c_api/c_api.cc
In file included from $SRC_DIR/src/c_api/../data/adapter.h:21,
                 from $SRC_DIR/src/c_api/c_api.cc:24:
$SRC_DIR/src/c_api/../data/array_interface.h:27:10: fatal error: cuda_fp16.h: No such file or directory
   27 | #include "cuda_fp16.h"  // for __half
      |          ^~~~~~~~~~~~~
compilation terminated.
```

.ci_support/linux_64_c_compiler_version11cuda_compilernvcccuda_compiler_version11.8cxx_compiler_version11.yaml
hcho3: Reverted

hcho3:
@jakirkham I reverted the lint-related changes. For now, we can ignore the lint error then?

jakirkham:
Yes please feel free to. Was planning to look more closely once we sorted out the build issues. Sorry if that was distracting

Do you have thoughts on the error we are seeing here ( https://github.com/conda-forge/xgboost-feedstock/pull/167#issuecomment-2246723367 )?

hcho3:
@jakirkham I added back `${CUDAToolkit_INCLUDE_DIRS}` to the include paths, but as a `SYSTEM`. This way, the CTK path will be searched after other include paths specified. 

> Additionally, system include directories are searched after normal include directories regardless of the order specified.

https://cmake.org/cmake/help/latest/command/target_include_directories.html

recipe/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
```diff
@@ -0,0 +1,63 @@
...
+ 
++  target_link_libraries(${target} PRIVATE CCCL::CCCL)
+   target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_CUDA=1)
+   target_include_directories(
+     ${target} PRIVATE
+-    ${xgboost_SOURCE_DIR}/gputreeshap
+-    ${xgboost_SOURCE_DIR}/rabit/include
++    ${xgboost_SOURCE_DIR}/gputreeshap)
++  target_include_directories(
++    ${target} SYSTEM PRIVATE
```
jakirkham, line 57: Also wonder if using `AFTER` might help (based on [this doc]( https://cmake.org/cmake/help/v3.30/command/target_include_directories.html )):

> By using `AFTER` or `BEFORE` explicitly, you can select between appending and prepending, independent of the default.

```suggestion
+    ${target} SYSTEM AFTER PRIVATE
```

jakirkham: Actually `SYSTEM` may already do this for us. Later in the same doc:

> Additionally, system include directories are searched after normal include directories regardless of the order specified.

hcho3: > system include directories are searched after normal include directories regardless of the order specified.

So not sure if `AFTER` would have an effect. I can give it a try.

jakirkham: Maybe we can try `AFTER` instead of `SYSTEM`?

Am guessing `SYSTEM` translates to `-isystem`, which might be resulting in some of the errors we are now seeing on CI

```suggestion
+    ${target} AFTER PRIVATE
```

jakirkham:
@conda-forge-admin , please re-render

recipe/meta.yaml
```diff
@@ -10,32 +11,36 @@ package:
...
     - 0003-Use-mingw-w64-path.patch
     - 0004-Undo-dmlc-xgboost-9436.patch
+    - 0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
 
 build:
   number: {{ build_number }}
-  skip: True  # [cuda_compiler_version == "11.2"]
+  {% if python is defined %}
+  skip: {{ python.split(".")[:2] != min_python.split(".")[:2] }}
+  {% endif %}
```
jakirkham, line 28: This checks that we only build Python packages once and that is for the minimum version that the packages support. It behaves similar to `noarch: python` that was added before. Though seems to be liked better by the linter

recipe/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
jakirkham: Tried adding this to see if it would help. Though happy to try other things if there are better options

recipe/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
```diff
@@ -0,0 +1,63 @@
+From 15b7012f54793f3cf807501453a61b6a6c1fb3c3 Mon Sep 17 00:00:00 2001
+From: Philip Hyunsu Cho <chohyu01@cs.washington.edu>
+Date: Tue, 23 Jul 2024 18:42:54 -0700
+Subject: [PATCH] [CMake] Explicitly link with CCCL (standalone or CTK)
+ (#10624)
+
+---
+ CMakeLists.txt    | 18 ++++++++++++++++++
+ cmake/Utils.cmake |  6 ++++--
```
jakirkham, line 9: Update the header relative to our changes

```suggestion
 cmake/Utils.cmake |  5 ++---
```

recipe/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
```diff
@@ -0,0 +1,63 @@
...
+ if(FORCE_COLORED_OUTPUT AND (CMAKE_GENERATOR STREQUAL "Ninja") AND
+diff --git a/cmake/Utils.cmake b/cmake/Utils.cmake
+index 8659df7c3..443b77c7b 100644
+--- a/cmake/Utils.cmake
++++ b/cmake/Utils.cmake
+@@ -108,11 +108,13 @@ function(xgboost_set_cuda_flags target)
+     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
+   endif()
+ 
++  target_link_libraries(${target} PRIVATE CCCL::CCCL)
```
jakirkham, line 49: Following [this pattern in cuCollections]( https://github.com/NVIDIA/cuCollections/blob/c05af722b45e8933eb079757dcb50f57fd255823/CMakeLists.txt#L86 )

```suggestion
+  target_link_libraries(${target} INTERFACE CCCL::CCCL CUDA::toolkit)
```

recipe/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
```diff
@@ -0,0 +1,63 @@
...
+   target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_CUDA=1)
+   target_include_directories(
+     ${target} PRIVATE
+-    ${xgboost_SOURCE_DIR}/gputreeshap
+-    ${xgboost_SOURCE_DIR}/rabit/include
++    ${xgboost_SOURCE_DIR}/gputreeshap)
++  target_include_directories(
++    ${target} AFTER PRIVATE
+     ${CUDAToolkit_INCLUDE_DIRS})
```
jakirkham, line 58: Think this is now handled by the `CUDA::toolkit` target above

```suggestion
-    ${CUDAToolkit_INCLUDE_DIRS})
```

jakirkham (commented):
Looks like we are still seeing the CCCL & CTK headers in the wrong order

Think these changes may help

recipe/patches/0004-Undo-dmlc-xgboost-9436.patch
```diff
@@ -25,31 +25,21 @@ index d6b861ab..ca0125cb 100644
...
- endfunction(xgboost_set_cuda_flags)
+@@ -240,7 +240,6 @@ macro(xgboost_target_link_libraries target)
  
- macro(xgboost_link_nccl target)
-@@ -280,7 +279,6 @@ macro(xgboost_target_link_libraries target)
- 
-   if (USE_CUDA)
+   if(USE_CUDA)
      xgboost_set_cuda_flags(${target})
 -    target_link_libraries(${target} PUBLIC CUDA::cudart_static)
```
jakirkham, line 39: Hyunsu, do you remember why we drop the `CUDA::cudart_static` linkage here?

Am wondering if it makes sense to keep it, but move it into `xgboost_set_cuda_flags`

recipe/patches/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
```diff
@@ -0,0 +1,60 @@
...
+diff --git a/cmake/Utils.cmake b/cmake/Utils.cmake
+index 8659df7..ec69278 100644
+--- a/cmake/Utils.cmake
++++ b/cmake/Utils.cmake
+@@ -108,12 +108,12 @@ function(xgboost_set_cuda_flags target)
+     target_compile_definitions(${target} PRIVATE -DXGBOOST_USE_NVTX=1)
+   endif()
+ 
++  target_link_libraries(${target} INTERFACE CCCL::CCCL)
++  target_link_libraries(${target} PUBLIC CUDA::cudart_static)
```
jakirkham, line 50: In terms of getting the include order right, this is doing what we want as [seen on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=986631&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=841356e0-85bb-57d8-dbbc-852e683d1642&l=1023 ) (as opposed to the reversed order seen previously)

```bash
-I$PREFIX/include ... -isystem /usr/local/cuda/targets/x86_64-linux/include
```

However am seeing some odd errors

```cpp
$PREFIX/include/cuda/std/__utility/pair.h(560): error: no operator "==" matches these operands
```

So likely there is still something off here

recipe/patches/0004-Undo-dmlc-xgboost-9436.patch
hcho3: I'm not quite sure. Let's see if we can get the build working with it.

recipe/patches/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK-10.patch
hcho3: Fix available at https://github.com/dmlc/xgboost/pull/10633

hcho3: Updated the patch.

hcho3: It turns out the particular error
```
$PREFIX/include/cuda/std/__utility/pair.h(560): error: no operator "==" matches these operands
```
was due to a bug in CCCL. So we should probably set a filter on the CCCL version to exclude the affected versions (2.4.0, 2.5.0).

jakirkham: Ah good catch. Thanks Hyunsu! 🙏

So we were picking up the newer CCCL. We just ran into a known bug

Here's hoping we are in the clear 🍀

hcho3: Yep, we are now using the latest CCCL for sure.

recipe/patches/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK.patch
```diff
@@ -0,0 +1,86 @@
...
+       return os;
+     }
++
++    XGBOOST_DEVICE inline bool operator==(Entry const& rhs) const {
++      return (rmin == rhs.rmin) && (rmax == rhs.rmax)
++             && (wmin == rhs.wmin) && (value == rhs.value);
++    }
++    XGBOOST_DEVICE inline bool operator!=(Entry const& rhs) const {
++      return !operator==(rhs);
++    }
```
jakirkham, line 80: Do we want to upstream this as well?

hcho3: I don't think we should upstream this patch. In fact, I was about to remove this workaround because it's due to a bug in CCCL. Instead, we should set a filter so that we don't use 2.4.0 or 2.5.0 version of CCCL.

hcho3: Another option is to retain this patch in the Conda recipe, until a new version of CCCL is released.

jakirkham:
Seeing this error [on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=987018&view=logs&j=517fe804-fa30-5dc2-1413-330699242c05&t=c10fa5f2-fdf6-5338-3bdb-c4bea7c23412&l=1042 );

```
FAILED: src/CMakeFiles/objxgboost.dir/tree/gpu_hist/evaluate_splits.cu.o 
$BUILD_PREFIX/bin/nvcc -forward-unknown-to-host-compiler -ccbin=$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-c++ -DDMLC_CORE_USE_CMAKE -DDMLC_LOG_CUSTOMIZE=1 -DDMLC_USE_CXX11=1 -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -DXGBOOST_BUILTIN_PREFETCH_PRESENT=1 -DXGBOOST_MM_PREFETCH_PRESENT=1 -DXGBOOST_USE_CUDA=1 -DXGBOOST_USE_NCCL=1 -D_MWAITXINTRIN_H_INCLUDED -D__USE_XOPEN2K8 -I$SRC_DIR/include -I$SRC_DIR/dmlc-core/include -I$SRC_DIR/rabit/include -I$SRC_DIR/gputreeshap -I$PREFIX/include -I$SRC_DIR/build-target/dmlc-core/include -isystem $BUILD_PREFIX/targets/x86_64-linux/include -O3 -DNDEBUG -std=c++17 "--generate-code=arch=compute_50,code=[sm_50]" "--generate-code=arch=compute_60,code=[sm_60]" "--generate-code=arch=compute_70,code=[sm_70]" "--generate-code=arch=compute_80,code=[sm_80]" "--generate-code=arch=compute_90,code=[sm_90]" "--generate-code=arch=compute_90,code=[compute_90]" -Xcompiler=-fPIC --expt-extended-lambda --expt-relaxed-constexpr -Xcompiler=-fopenmp -Xfatbin=-compress-all --default-stream per-thread -lineinfo -MD -MT src/CMakeFiles/objxgboost.dir/tree/gpu_hist/evaluate_splits.cu.o -MF src/CMakeFiles/objxgboost.dir/tree/gpu_hist/evaluate_splits.cu.o.d -x cu -c $SRC_DIR/src/tree/gpu_hist/evaluate_splits.cu -o src/CMakeFiles/objxgboost.dir/tree/gpu_hist/evaluate_splits.cu.o
$SRC_DIR/src/tree/gpu_hist/../updater_gpu_common.cuh(57): error: identifier "FLT_MAX" is undefined
```

Guessing we want to apply this patch: https://github.com/dmlc/xgboost/pull/10574

recipe/patches/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK.patch
jakirkham: Ah ok. Am now seeing this context in this thread: https://github.com/dmlc/xgboost/pull/10633#discussion_r1690225473

Yep am happy with either option. Will include the other option as a review for comparison/discussion

recipe/meta.yaml
```diff
@@ -10,32 +11,36 @@ package:
...
+    - {{ compiler('m2w64_cxx') }}  # [win]
+    - {{ compiler('cuda') }}       # [cuda_compiler != "None" and cuda_compiler_version != "None"]
     - {{ stdlib('c') }}
     - cmake
     - llvm-openmp  # [osx]
     - libgomp      # [linux]
     - ninja
   host:
+    - cccl         # [cuda_compiler != "None"]
     - nccl         # [linux and cuda_compiler != "None"]
```
jakirkham, line 44: This will exclude both of the problematic `cccl` versions as discussed above. If we need to exclude more, we can modify this further

```suggestion
    - cccl !=2.4.0,!=2.5.0  # [cuda_compiler != "None"]
    - nccl                  # [linux and cuda_compiler != "None"]
```

jakirkham (commented):
Including the option to skip CCCL versions with a known bug encountered in our builds as discussed in this thread: https://github.com/conda-forge/xgboost-feedstock/pull/167#discussion_r1690279847

jakirkham:
Am seeing this error in the [Windows CUDA 11.8 builds]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=987018&view=logs&j=e5cdccbf-4751-5a24-7406-185c9d30d021&t=0c979b57-e97b-5e0a-363d-f964823a2ab0&l=6207 ):

```
Internal error: assertion failed at: "overload.c", line 28946 in deduce_class_template_args


1 catastrophic error detected in the compilation of "D:/bld/xgboost-split_1721846306365/work/src/data/ellpack_page_source.cu".
Compilation aborted.

ellpack_page_source.cu

nvcc error   : 'cudafe++' died with status 0xC0000409 
```

Is this related to the patch we are discussing above ( https://github.com/conda-forge/xgboost-feedstock/pull/167#discussion_r1690279847 )? If so, maybe this is a good reason to consider dropping it and restricting CCCL instead

hcho3:
> Is this related to the patch we are discussing above ( https://github.com/conda-forge/xgboost-feedstock/pull/167#discussion_r1690279847 )? If so, maybe this is a good reason to consider dropping it and restricting CCCL instead

Not sure, we can first try restricting CCCL and see if the internal error goes away.

recipe/patches/0006-Partial-fix-for-CTK-12.5-10574.patch
```diff
@@ -0,0 +1,65 @@
...
+Subject: [PATCH] Partial fix for CTK 12.5 (#10574)
+
+---
+ src/tree/updater_gpu_common.cuh | 21 ++++++++-------------
+ src/tree/updater_gpu_hist.cu    |  1 +
+ 2 files changed, 9 insertions(+), 13 deletions(-)
+
+diff --git a/src/tree/updater_gpu_common.cuh b/src/tree/updater_gpu_common.cuh
+index 1c3e6a552..5d999d6d6 100644
+--- a/src/tree/updater_gpu_common.cuh
```
trivialfis, line 13: We can backport this to 2.1

recipe/patches/0004-Undo-dmlc-xgboost-9436.patch
```diff
@@ -1,20 +1,20 @@
-From 667a855dfa024ea392e60fa7f352a0781cb7f1e0 Mon Sep 17 00:00:00 2001
-From: Hyunsu Cho <chohyu01@cs.washington.edu>
-Date: Thu, 2 Nov 2023 08:12:43 -0700
-Subject: [PATCH] Undo dmlc/xgboost#9436
+From 3660a4216763d02d48cb1db63f5fee7ad02ad7d4 Mon Sep 17 00:00:00 2001
+From: Hyunsu Cho <phcho@nvidia.com>
```
trivialfis, line 2: Since the original patch was intended for conda packaging, maybe we can revert it in XGBoost if it's no longer needed?

jakirkham: Well there are two things going on

1. Reverting [this RMM workaround]( https://github.com/dmlc/xgboost/blob/411c8466bd383662a448a6a6be585c1b12b8c2e2/CMakeLists.txt#L320-L327 ). AIUI this is needed until RMM fixes issue: https://github.com/rapidsai/rmm/issues/1320
2. Moving `CUDA::cudart_static` into `xgboost_set_cuda_flags`, which is being done in PR: https://github.com/dmlc/xgboost/pull/10633

We can probably cleanup the patches here to make these things clearer (and also make dropping them easier in the future)

Though wanted to wait until the logic here settled (namely we got a green CI) before embarking on that. Otherwise the diff just gets messier and harder to follow while we are still debugging issues

trivialfis: Thank you for the explanation, it's really helpful!

jakirkham:
The good new is it appears most of the errors we saw before are clearing out! 🥳 Thanks for all the patches to make that happen ❤️

The bad news is we are still seeing the new Windows compilation issue mentioned above: https://github.com/conda-forge/xgboost-feedstock/pull/167#issuecomment-2248731210

recipe/patches/0004-Undo-dmlc-xgboost-9436.patch
jakirkham: Happy to help 🙂

Know looking at a diff of a diff is horrendous. So have just been viewing each file directly in its current state

Am encouraged by the work Hyunsu and you have been doing. Think we will be able to start cutting these down in future releases

jakirkham:
@conda-forge-admin , please re-render

recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+c_compiler:                    # [win]
+- vs2022                       # [win]
+cxx_compiler:                  # [win]
+- vs2022                       # [win]
```
jakirkham, line 4: Windows uses Visual Studio 2019 by default

In case we are seeing a bug in Visual Studio 2019, this upgrades us to Visual Studio 2022

jakirkham:
The Windows CUDA 12 build has this error [on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=987203&view=logs&j=5be07ae1-d8ba-5406-47b6-8e3a3a12f825&t=c94d5231-acfc-5e67-dafc-77bdd383ead4&l=2562 ):

```
  D:\bld\xgboost-split_1721864376122\_build_env\Library\include\crt/host_config.h(153):
  fatal error C1189: #error: -- unsupported Microsoft Visual Studio version!
  Only the versions between 2017 and 2022 (inclusive) are supported! The nvcc
  flag '-allow-unsupported-compiler' can be used to override this version
  check; however, using an unsupported host compiler may cause compilation
  failure or incorrect run time execution.  Use at your own risk.
```

This is odd as it says 2022 should be supported and that is what we are using


Will wait and see if this helps Windows CUDA 11.8 before spending more time on this angle

jakirkham:
@conda-forge-admin , please re-render

recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+c_compiler:                    # [win]
+- vs2017                       # [win]
+cxx_compiler:                  # [win]
+- vs2017                       # [win]
```
jakirkham, line 4: Trying downgrading to VS 2017. Not ideal as the default is VS 2019. However if we need it, this is an option

jakirkham: Looks like we may not have VS 2017 on the Azure images any more. So that doesn't look like an option. Will revert back to the defaults

hcho3:
I think the error comes from a compiler bug from NVCC 11.8. I was able to repro the same error locally with NVCC 11.8 + MSVC 2022.

Let's consider dropping 11.8 for Windows then?

jakirkham:
@conda-forge-admin , please re-render

jakirkham:
@conda-forge-admin , please re-render

recipe/bld.bat
```diff
@@ -1,20 +1,36 @@
...
+:: This happens in spite of the fact that VS 2019 & 2022 default to C++14.
+:: To make sure `__cplusplus` matches, we pass an additional flag to VC.
+:: Ideally other tooling would do this for us, for now we do it ourselves.
+::
+:: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-160
+:: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170
+:: https://github.com/NVIDIA/cccl/blob/82a3ed0282893d6316abde04394f59a9a37a3747/libcudacxx/include/nv/target#L27-L30
+:: https://github.com/NVIDIA/cutlass/issues/1403#issuecomment-2130383392
+:: https://gitlab.kitware.com/cmake/cmake/-/issues/18837
+set "CXXFLAGS=%CXXFLAGS% /Zc:__cplusplus"
```
jakirkham, line 18: Ok did a bit of a deep dive and have a theory of what is going on. Tried to explain this in the comment above

Basically some C++ detection logic in libraries ([like CCCL]( https://github.com/NVIDIA/cccl/blob/82a3ed0282893d6316abde04394f59a9a37a3747/libcudacxx/include/nv/target#L27-L30 )) doesn't work correct with Visual Studio due to its default behavior. Have added a flag to Visual Studio to make it conform more to typical C++ compiler behavior (setting `__cplusplus` correctly)

Based on similar experiences from Visual Studio users with CUTLASS ( https://github.com/NVIDIA/cutlass/issues/1403#issuecomment-2130383392 ), who saw a similar error to us and [fixed it with this flag]( https://github.com/NVIDIA/cutlass/blob/56b46e2d13875b46b8f6a03f9f5ac91e2bfdc01a/CMakeLists.txt#L442-L452 ), am hopeful this will help in our case too

Ideally other tooling would handle this for us, but it appears [CMake hasn't done this yet]( https://gitlab.kitware.com/cmake/cmake/-/issues/18837 ) and Visual Studio continues to require users to add this flag if they want a correct `__cplusplus` value. So for now adding to this build script. If this works, we may want to add this to XGBoost as well

jakirkham:
> Let's consider dropping 11.8 for Windows then?

If the latest fix change doesn't work, then we can turn this into an issue and skip it

recipe/bld.bat
jakirkham: Have raised an upstream CCCL issue: https://github.com/NVIDIA/cccl/issues/2072

Though after looking more closely at the code there, they are probably ok with their current approach. This would be a potential simplification for them

recipe/meta.yaml
```diff
@@ -10,33 +11,38 @@ package:
...
-    - 0004-Undo-dmlc-xgboost-9436.patch
+    - patches/0001-Force-endian-flag-in-cross-compilation-mode.patch  # [arm64 or aarch64 or ppc64le]
+    - patches/0002-Enable-latest-libcxx-on-MacOS.patch  # [osx]
+    - patches/0003-Use-mingw-w64-path.patch
+    - patches/0004-Undo-dmlc-xgboost-9436.patch
+    - patches/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK.patch
+    - patches/0006-Partial-fix-for-CTK-12.5-10574.patch
 
 build:
   number: {{ build_number }}
```
jakirkham, line 26:
```suggestion
  number: {{ build_number }}
  # Windows CUDA 11.8 fails due to a compilation error
  # xref: https://github.com/conda-forge/xgboost-feedstock/issues/173
  skip: True  # [win and cuda_compiler_version == "11.8"]
```

jakirkham:
@conda-forge-admin , please re-render

.ci_support/win_64_cuda_compilernvcccuda_compiler_version11.8python3.8.____cpython.yaml
```diff
@@ -17,11 +17,11 @@ m2w64_c_compiler:
...
-  r-base:
+  python:
     min_pin: x.x
     max_pin: x.x
-r_base:
-- '4.1'
+python:
+- 3.8.* *_cpython
 target_platform:
 - win-64
```
jakirkham, line 26: This is the Windows CUDA 11.8 file. It should be removed by the `skip`, but it is not for some reason. Will need to take a closer look

Note part of the issue here is even if the build doesn't create any packages simply downloading the CTK can take over 20mins. So will want to make sure this job is properly removed so it isn't wasting CI cycles

recipe/meta.yaml
```diff
@@ -10,33 +11,41 @@ package:
...
-    - 0001-Force-endian-flag-in-cross-compilation-mode.patch  # [arm64 or aarch64 or ppc64le]
-    - 0002-Enable-latest-libcxx-on-MacOS.patch  # [osx]
-    - 0003-Use-mingw-w64-path.patch
-    - 0004-Undo-dmlc-xgboost-9436.patch
+    - patches/0001-Force-endian-flag-in-cross-compilation-mode.patch  # [arm64 or aarch64 or ppc64le]
+    - patches/0002-Enable-latest-libcxx-on-MacOS.patch  # [osx]
+    - patches/0003-Use-mingw-w64-path.patch
+    - patches/0004-Undo-dmlc-xgboost-9436.patch
+    - patches/0005-CMake-Explicitly-link-with-CCCL-standalone-or-CTK.patch
+    - patches/0006-Partial-fix-for-CTK-12.5-10574.patch
```
jakirkham, line 23: @hcho3 when you have a chance, could you please tidy up these patches (possibly using [`git-format-patch`]( https://git-scm.com/docs/git-format-patch ))?

As illustrated by the conversation with Jiaming above ( https://github.com/conda-forge/xgboost-feedstock/pull/167#discussion_r1690366111 ), this would likely make these now (in many cases simplified patches) easier to follow

.ci_support/win_64_cuda_compilernvcccuda_compiler_version11.8python3.8.____cpython.yaml
hcho3: @jakirkham I added back `noarch: python` and re-rendered the recipe. It looks like our replacement for `noarch: python` somehow broke the selector logic. Once I added back `noarch: python` to `meta.yaml`, the build for Windows CUDA 11.8 is finally removed.

recipe/meta.yaml
hcho3: Done. I left the first three patches, since some of them are applied to only specific system targets.

recipe/meta.yaml
```diff
@@ -24,12 +24,10 @@ source:
 
 build:
   number: {{ build_number }}
+  noarch: python
   # Windows CUDA 11.8 fails due to a compilation error
   # xref: https://github.com/conda-forge/xgboost-feedstock/issues/173
   skip: True  # [win and cuda_compiler_version == "11.8"]
-  {% if python is defined %}
-  skip: {{ python.split(".")[:2] != min_python.split(".")[:2] }}
-  {% endif %}
```
jakirkham: Good intuition Hyunsu! 🙏

Think we can fix this up to handle both cases then and fix the linter

Though ok doing this as a follow up

jakirkham (approved):

jakirkham:
Thanks Hyunsu

Looks good. Please feel free to add the `automerge` label when you are ready

The bot will then merge when CI completes. If that doesn't happen for some reason, it will leave a comment letting us know why

jakirkham:
Thanks Hyunsu! 🙏

trivialfis:
Hi @jakirkham @hcho3 is it normal to have this under a CUDA 12.5 environment:
```
  + libxgboost                                   2.1.0  cuda120_h9dfd3e9_1                 conda-forge/linux-64       93MB
  + xgboost                                      2.1.0  cuda118_pyh98e67c5_1               conda-forge/noarch         17kB
```
I'm updating an environment that has XGBoost as a dependency. The above two lines are part of the mamba output, and it seems a bit weird to me to have mismatched CUDA versions.

jakirkham:
@trivialfis this is a good question

AIUI in terms of CUDA usage, only `libxgboost` has CUDA bits. So `xgboost` doesn't have CUDA bits (as it is pure Python)

That said, it certainly is confusing to have `xgboost` mislabeled in this way. Also users that constrain `xgboost` based on CUDA version may also expect this to carry through to `libxgboost`

So agree we should fix this. Though think the risk is relatively low

jakirkham:
> Second question, since all C++ includes are distributed, maybe it makes sense to distribute the CMake config files as well?

Also sorry for not answering the CMake question earlier

Thank you for opening a new issue to discuss: https://github.com/conda-forge/xgboost-feedstock/issues/189

Hopefully the answers there are helpful

PR #194, opened 2024-09-28
Mark Python wheels as any platform

xhochy:
Checklist


xhochy:
Logs look fine: `filename=xgboost-2.1.1-py3-none-any.whl`

jakirkham:
Thanks Uwe! 🙏

@hcho3 would you be able to help upstream this change?

jakirkham:
How about extending the test that was added in PR ( https://github.com/dmlc/xgboost/pull/10755 )?

hcho3:
@jakirkham Not sure what you mean by that. To clarify my remark earlier: in the mainline XGBoost, the pip wheels must have system-specific tags, and cannot be just `py3-none-any`.

jakirkham:
Talked with Hyunsu offline about this. Summarizing below. Please correct me if I misunderstood anything @hcho3

Think I was just confused about how [this patch removing NCCL worked]( https://github.com/dmlc/xgboost/pull/10755/files#diff-5c8b5efc69e22e7d6fe244c3741e8a43bc27b1c70e904ca9e6b640506d1ee07d ). Thought that was being used to test a particular build of XGBoost. So thought a similar test could be added for no platform tags

However that doesn't appear to be the case. Instead the patch removing NCCL is used in upstream XGBoost's CPU wheel builds. So isn't actually a test

Agree if upstream is packaging `libxgboost` in each `xgboost` wheel, then there need to be system tags upstream

PR #196, opened 2024-10-31
Rebuild for r-base 4.4 and UCRT mingw toolchain

danielnachun:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

Closes https://github.com/conda-forge/xgboost-feedstock/pull/177


danielnachun:
As suggested in https://github.com/conda-forge/xgboost-feedstock/pull/177#issuecomment-2448990783, I have created a new PR with some fixes I found in local testing.  The R library now compiles successfully but fails to load due to a seemingly unrelated error with the `stats` package (reformatted for readbility):
```
  out: make: Nothing to be done for 'all'.
  During startup - Warning message:
  package 'stats' in options("defaultPackages") was not found
  Error in inDL(x, as.logical(local), as.logical(now), ...) :
    unable to load shared object 'D:/bld/xgboost-split_1730351293412/_h_env/lib/R/library/stats/libs/x64/stats.dll':
    LoadLibrary failure:  The specified procedure could not be found.

  Calls: <Anonymous> ...  asNamespace -> loadNamespace -> library.dynam ->
  dyn.load -> inDL

  Execution halted
  , err: During startup - Warning message:
  package 'stats' in options("defaultPackages") was not found

  * installing to library
  'D:/bld/xgboost-split_1730351293412/_h_env/lib/R/library'
  * installing *source* package 'xgboost' ...
  ** using staged installation
  ** libs
  installing to
  D:/bld/xgboost-split_1730351293412/_h_env/lib/R/library/00LOCK-R-package/00new/xgboost/libs/x64
  ** R
  ** data
  ** demo
  ** inst
  ** byte-compile and prepare package for lazy loading
  ERROR: lazy loading failed for package 'xgboost'
  * removing
  'D:/bld/xgboost-split_1730351293412/_h_env/lib/R/library/xgboost'

**Comment by danielnachun** [comment]
@h-vetinari I was able to fix the R package issues - I just had to tweak `PATH` on Windows.  However, the R package is now failing its test on the Windows CUDA build with no clear error.  I'm wondering if we can just disable the Windows CUDA build for now.  This package is blocking our R 4.4 migration and having the CPU package at least will let us finish that migration.

**Comment by h-vetinari** [comment]
> I was able to fix the R package issues

Nice!

> However, the R package is now failing its test on the Windows CUDA build with no clear error. I'm wondering if we can just disable the Windows CUDA build for now.

I'm neither part of @conda-forge/xgboost nor @conda-forge/r, so it's better to discuss with those maintainers. That said, I have nothing against unblocking the migration based on the CPU-only builds.

**Comment by h-vetinari** [comment]
Actually, it's possible that the package is fully functional, and that the test just fails because some part of xgboost wants to load `cuda.dll`, fails, and produces garbage (e.g. zeros, nulls) instead of an error. The fact that the losses in the failing test look like this
```
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
```
possibly point in that direction (i.e. the solver might fail completely and we just get the norm of the original vector).

**Comment by danielnachun** [comment]
I didn't notice those were all repeated before, but it seems like is also true even on a Linux non-CUDA test:
```
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
[1]	train-logloss:0.233156 
[2]	train-logloss:0.136542 
```
Perhaps our test isn't working how we expect it to?

**Comment by danielnachun** [comment]
I have a local Windows machine where I'm going to test this to see if I can get some more informative output - annoyingly there are more than a few cases where the output is only visible with a GUI.

**Comment by h-vetinari** [comment]
Can you try rebasing off of main for the most recent CUDA changes?

**Comment by danielnachun** [comment]
Windows passed at last!  I'm not sure what was changed with CUDA but we should be all set now.  If one of the maintainers for this feedstock can merge this when it's done (or just set the `automerge` label), we should finally be able to merge this.  Huge thanks to @h-vetinari for help with this!

**Comment by jakirkham** [comment]
There were a couple XGBoost patch releases. Perhaps one of those helped?

File: recipe/install-r-xgboost.bat [hunk 1]
```diff
@@ -2,10 +2,14 @@
 mkdir "%SRC_DIR%"\build-r
 pushd "%SRC_DIR%"\build-r
 
+set PATH=%LIBRARY_PREFIX%\bin;%PATH%
```
jakirkham, line 5: Is this still needed? Would think `%LIBRARY_PREFIX%\bin` is already in `%PATH%`

recipe/install-r-xgboost.bat
```diff
@@ -2,10 +2,14 @@
...
+
 cmake -G "Ninja" ^
     -DCMAKE_BUILD_TYPE:STRING="Release" ^
     -DCMAKE_INSTALL_PREFIX="%LIBRARY_PREFIX%" ^
     -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=ON ^
+    -DDLLTOOL_EXE=%BUILD_PREFIX%\Library\bin\%DLLTOOL% ^
```
jakirkham, line 11: Also is this still needed? Would think `%BUILD_PREFIX%\Library\bin` is in `%PATH%`. So would think `% DLLTOOL%` would get picked up by the normal search process

Though please let me know if there is something more to this

recipe/install-r-xgboost.bat
```diff
@@ -2,10 +2,14 @@
...
+    -DLIBR_CORE_LIBRARY=%PREFIX%\lib\R\bin\x64\R.lib ^
```
jakirkham, line 12: Could you please add a comment here? It would help future readers understand when this may or may not be needed (particularly for other R recipes)

recipe/meta.yaml
```diff
@@ -35,10 +35,13 @@ requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
-    - {{ compiler('m2w64_c') }}    # [win]
-    - {{ compiler('m2w64_cxx') }}  # [win]
-    - {{ compiler('cuda') }}       # [cuda_compiler != "None" and cuda_compiler_version != "None"]
+    - {{ compiler('cuda') }}           # [cuda_compiler != "None" and cuda_compiler_version != "None"]
+    - {{ compiler('m2w64_c') }}        # [win]
+    - {{ compiler('m2w64_cxx') }}      # [win]
+    - {{ compiler('m2w64_fortran') }}  # [win]
```
jakirkham, line 41: Do we still need MinGW Fortran or can it be dropped?

recipe/meta.yaml
```diff
@@ -210,8 +213,10 @@ outputs:
     script: install-r-xgboost.sh  # [not win]
     script: install-r-xgboost.bat  # [win]
     build:
-      skip: true  # [r_base in ("4.1", "4.2") or win]
+      skip: true  # [r_base in ("4.1", "4.2")]
       string: {{ string_prefix }}_r{{ r_base | replace('.', '') }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}
+      ignore_run_exports_from:             # [win]
+        - {{ compiler('m2w64_fortran') }}  # [win]
```
jakirkham, line 219: If we dropped MinGW Fortran above, this also can be dropped

recipe/meta.yaml
```diff
@@ -220,28 +225,31 @@ outputs:
...
-        - {{ compiler('m2w64_c') }}    # [win]
-        - {{ compiler('m2w64_cxx') }}  # [win]
-        - {{ compiler('c') }}
-        - {{ compiler('cxx') }}
-        - {{ stdlib('c') }}
-        - llvm-openmp                  # [osx]
-        - libgomp                      # [linux]
+        - {{ compiler('m2w64_c') }}        # [win]
+        - {{ compiler('m2w64_cxx') }}      # [win]
+        - {{ compiler('m2w64_fortran') }}  # [win]
```
jakirkham, line 230: Same question about Fortran

recipe/meta.yaml
```diff
@@ -220,28 +225,31 @@ outputs:
...
+        - {{ stdlib('m2w64_c') }}          # [win]
+        - {{ compiler('c') }}              # [unix]
+        - {{ compiler('cxx') }}            # [unix]
+        - {{ stdlib('c') }}                # [unix]
```
jakirkham, line 234: Maybe we can order these to match the top-level list of compilers. Think this could help when making sure they have similar things

recipe/meta.yaml
```diff
@@ -220,28 +225,31 @@ outputs:
...
+        - r-base                           # [build_platform != target_platform]
+        - r-matrix                         # [build_platform != target_platform]
+        - r-data.table                     # [build_platform != target_platform]
+        - r-magrittr                       # [build_platform != target_platform]
+        - r-jsonlite                       # [build_platform != target_platform]
+        - r-knitr                          # [build_platform != target_platform]
       host:
         - {{ pin_subpackage('_r-xgboost-mutex', exact=True) }}
         - {{ pin_subpackage('libxgboost', max_pin='x.x.x') }}
+        - libgomp                          # [not osx]
```
jakirkham, line 252: This is already added via `libgomp`'s `run_exports/strong` and it is in the top-level. So let's drop this

```suggestion
```

recipe/meta.yaml
```diff
@@ -252,6 +260,7 @@ outputs:
       run:
         - {{ pin_subpackage('_r-xgboost-mutex', exact=True) }}
         - {{ pin_subpackage('libxgboost', max_pin='x.x.x') }}
+        - libgomp                          # [win]
```
jakirkham, line 263: This is already added via `libgomp`'s `run_exports/strong` and it is in the top-level. So let's drop this

```suggestion
```

jakirkham (commented):
Thanks Daniel and Axel! 🙏

Had some questions below

recipe/install-r-xgboost.bat
h-vetinari: FYI, it's not possible to add comments within the command that's broken across several lines. It can only be done before/after

jakirkham: Adding before would be ok

recipe/install-r-xgboost.bat
danielnachun: Yes, this fix is needed because `%BUILD_PREFIX%\bin` comes before `%LIBRARY_PREFIX%\bin%, and this breaks the build. So prepending `%LIBRARY_PREFIX%\bin` seems to be needed to guarantee that for executables found in both prefixes, the one in `%LIBRARY_PREFIX%\bin` is preferred.

danielnachun: Having said that, it is worth trying to remove it now and see if it's still needed.

recipe/install-r-xgboost.bat
danielnachun: I believe this was manually added because `cmake` was hard-coding an executable name for %DLLTOOL% that didn't exist.  These issues seem to be the result of changes to executable names with the UCRT migration.  We can also try removing this to see if it's still needed.

recipe/install-r-xgboost.bat
danielnachun: I also had to add this because `cmake` was choosing the wrong value here.  We can try removing it, and I'll at least comment explaining this if it is still needed.

recipe/meta.yaml
danielnachun: This should probably dropped - the only reason it might be needed is if `cmake` checks for a Fortran compiler and fails if it doesn't find one.

recipe/meta.yaml
danielnachun: Agreed, this recipe is hard enough to follow as is so any readability improvements will help.

recipe/install-r-xgboost.bat
jakirkham: Sounds good let's try removing

If it is still needed, including a code comment along the lines of what you have written would be helpful for future readers

recipe/install-r-xgboost.bat
jakirkham: And if it is still needed, a code comment would be much appreciated

recipe/meta.yaml
jakirkham: This is probably leftover from my past attempts to get things working ( https://github.com/conda-forge/xgboost-feedstock/commit/4ccf854fe2c91f69b379dbcdb215eb0035d82c09 ). Likely it is safe to drop

recipe/meta.yaml
danielnachun: If `libgomp` is in `build` requirements, will that still give a `run_export`?  I thought that only happened with `host`, and the top level `libgomp` is in `build`, not `host`.

danielnachun:
Now that we have a working starting point, I will try systematically disabling the various workarounds one at a time, adding a comment for each one we need to keep.

recipe/meta.yaml
jakirkham: Yes, adding a package in `requirements/build` will ensure its `run_exports/strong` requirements will be applied in `requirements/host` and `requirements/run`. From the [conda-build docs]( 
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements ):

> With `run_exports` (new in conda-build 3) such a runtime requirement can be implicitly added by host requirements (e.g. libpng exports libpng), and with `run_exports/strong` even by build requirements (e.g. GCC exports libgcc).

`libgomp` specifically adds [`run_exports/strong` on `_openmp_mutex `]( https://github.com/conda-forge/ctng-compilers-feedstock/blob/09c04e6458ab4d79193b4c20a6330e84540108ec/recipe/meta.yaml#L507-L509 ), which ensures `_openmp_mutex ` is automatically added to `requirements/host` and `requirements/run`. `_openmp_mutex` [depends on `libgomp` at runtime ]( https://github.com/conda-forge/ctng-compilers-feedstock/blob/09c04e6458ab4d79193b4c20a6330e84540108ec/recipe/meta.yaml#L579-L581 ). Meaning anywhere `_openmp_mutex` is installed `libgomp` is too, which means anything that had `libgomp` in `requirements/build`. The `_openmp_mutex` also [tests the `libgomp` SO is available and usable]( https://github.com/conda-forge/ctng-compilers-feedstock/blob/09c04e6458ab4d79193b4c20a6330e84540108ec/recipe/meta.yaml#L586-L589 )

Hence [the recommended usage of OpenMP in conda-forge]( https://conda-forge.org/docs/maintainer/knowledge_base/#openmp ) is to only added OpenMP dependencies in `requirements/build` and let `run_exports/strong` of these dependencies to fill in the rest of `requirements` automatically

Should add XGBoost specifically [tests that all linkages are satisfied]( https://github.com/conda-forge/xgboost-feedstock/blob/0cd682cc56d8b5c44b157571fe9587028c0c94a1/conda-forge.yml#L7-L8 ). So if we linked to OpenMP and the needed libraries are not present, the build would error. Note the builds pass following conda-forge OpenMP recommendations meaning this was working as expected

danielnachun: Thanks for the detailed explanation.  We had some issues with a few R packages where `libgomp` wasn't being included in `run_exports` and failed linkage checks on Windows.  I need to revisit those with this information - it's possible we were actually hitting a bug with `run_export` propagation rather than simply missing a dependency.  I'll follow up on this on Zulip if the behavior of `libgomp` is inconsistent with what is expected for `run_exports/strong`.

jakirkham: Meaning this list: https://github.com/conda-forge/r-base-feedstock/issues/325#issuecomment-2342479080 ?

Scanned through that list and left a few comments in associated PRs and left a few notes

This one ( https://github.com/conda-forge/r-idefix-feedstock/pull/1 ) is notable as it wasn't doing the right thing with OpenMP to start with. Think fixing that would be the best place to focus effort

The others are pretty simple cleanups after that (mainly deleting the extra OpenMP lines)

danielnachun: Yes, that list was not all of them but it did cover many.  I'm now realizing the issue might actually have been that `libgomp` had been a Linux-only dependency and needed to be `not osx` instead.  I'll try to change those to just have `libgomp` where it's needed.

danielnachun:
All of the workarounds in `build.bat` from before were still needed so I have now added comments for them.

Unfortunately, when I only have `libgomp` in build requirements, we still get this linkage audit failure on Windows:
```
  ERROR (r-xgboost,lib/R/library/xgboost/libs/x64/xgboost.dll): Needed DSO Library/bin/libgomp-1.dll found in ['conda-forge/win-64::libgomp==14.2.0=h1383e82_1']
  ERROR (r-xgboost,lib/R/library/xgboost/libs/x64/xgboost.dll): .. but ['conda-forge/win-64::libgomp==14.2.0=h1383e82_1'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

I'm now checking what happens if I put `libgomp` into host requirements, but even if that is sufficient that does seem to indicate something is not behaving as expected with `run_exports`.

danielnachun:
The only change now was just removing `libgomp` from the run requirements for `r-xgboost` on Windows.  I tried putting it in host, and the same error still occurred.  So please feel free to take a look at these logs now - I won't push any more changes for now until we can understand why `conda-build` thinks that `libgomp` isn't in `reqs/run` even though it is now included in both build and host requirements.

As of now the only way this has been fixed in the previous successful has been to manually add `libgomp` to the run requirements on Windows.  And confusingly, this isn't an issue on Linux and as you said above the logic is supposed to be same for `libgomp` for both OSes.

danielnachun:
Out of curiosity I tried removing `libgomp` from the run requirements for Windows for this R feedstock: https://github.com/conda-forge/r-fedmatch-feedstock/pull/2, and the same linkage audit failure occurs.  This would seem to indicate that this is not specific to this package.

If there's no obvious fix here, I wonder if it would make more sense to add `libgomp` to the run requirements here on Windows so we can merge this and unblock the R 4.4 migration for the dependents of this package.  Then we could iterate on this underlying issue in https://github.com/conda-forge/r-fedmatch-feedstock/pull/2 more quickly since that PR can finish a lot faster than this one.

danielnachun:
@jakirkham any thoughts on how to move forward with this?

bgruening:
friendly ping @jakirkham :)

theAeon:
just for the record i'm running into this same link issue over at https://github.com/conda-forge/r-rcppplanc-feedstock#1 so I think its possible something's just broken w/ libgomp

danielnachun:
@theAeon Agreed, I'm fairly sure the run export for `libgomp` isn't working as expected on Windows.  For now I think manually adding the extra run dependency is sufficient for completing the R 4.4 migration but eventually we will want to fix this permanently.

theAeon:
Will have more of an idea after my builds run-build queue is packed for some reason 😅

theAeon:
Hm. Still doesn't like me. Did yall need it in build, host, AND run?

isuruf:
This is an issue with our gcc windows toolchain. Please add `libgomp  # [win]` to `run` for now

recipe/meta.yaml
```diff
@@ -28,13 +28,15 @@ requirements:
...
+    - {{ compiler('cuda') }}           # [cuda_compiler != "None" and cuda_compiler_version != "None"]
+    - {{ compiler('m2w64_c') }}        # [win]
+    - {{ compiler('m2w64_cxx') }}      # [win]
     - {{ stdlib('c') }}
+    - {{ stdlib('m2w64_c') }}          # [win]
+    - m2-sed                           # [win]
     - cmake
     - llvm-openmp  # [osx]
-    - libgomp      # [linux]
+    - libgomp      # [not osx]
```
jakirkham, line 39:
```suggestion
    - libgomp      # [linux or win]
```

recipe/meta.yaml
```diff
@@ -66,7 +68,7 @@ outputs:
         - cmake
         - make
         - llvm-openmp  # [osx]
-        - libgomp      # [linux]
+        - libgomp      # [not osx]
```
jakirkham, line 71:
```suggestion
        - libgomp      # [linux or win]
```

recipe/meta.yaml
```diff
@@ -213,28 +215,30 @@ outputs:
...
-        - llvm-openmp                  # [osx]
-        - libgomp                      # [linux]
+        - {{ compiler('c') }}              # [unix]
+        - {{ compiler('cxx') }}            # [unix]
+        - {{ compiler('m2w64_c') }}        # [win]
+        - {{ compiler('m2w64_cxx') }}      # [win]
+        - {{ stdlib('c') }}                # [unix]
+        - {{ stdlib('m2w64_c') }}          # [win]
+        - llvm-openmp                      # [osx]
+        - libgomp                          # [not osx]
```
jakirkham, line 225:
```suggestion
        - libgomp                          # [linux or win]
```

recipe/meta.yaml
```diff
@@ -213,28 +215,30 @@ outputs:
...
+        - r-base                           # [build_platform != target_platform]
+        - r-matrix                         # [build_platform != target_platform]
+        - r-data.table                     # [build_platform != target_platform]
+        - r-magrittr                       # [build_platform != target_platform]
+        - r-jsonlite                       # [build_platform != target_platform]
+        - r-knitr                          # [build_platform != target_platform]
       host:
         - {{ pin_subpackage('_r-xgboost-mutex', exact=True) }}
         - {{ pin_subpackage('libxgboost', max_pin='x.x.x') }}
+        - libgomp      # [not osx]
```
jakirkham, line 241: Only Windows seems to need this here. For Linux `run_exports` works as intended

```suggestion
        - libgomp      # [win]
```

recipe/meta.yaml
```diff
@@ -66,7 +68,7 @@ outputs:
         - cmake
         - make
         - llvm-openmp  # [osx]
-        - libgomp      # [linux]
+        - libgomp      # [not osx]
       host:
         - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler != "None"]
         - nccl                                      # [linux and cuda_compiler != "None"]
```
jakirkham, line 74: Sounds like we will want this in `run` as well

```suggestion
        - libgomp                                   # [win]
```

jakirkham:
> This is an issue with our gcc windows toolchain. Please add `libgomp  # [win]` to run for now

Thanks Isuru! 🙏

Is there an associated GitHub issue?

jakirkham (commented):
Thanks Daniel! Appreciate your persistence here 🙏

Have added a few suggestions below to hopefully keep the Windows workaround traceable so we can cleanly back it out once it is fixed upstream

Sorry for the slow replies here. Have unfortunately not had lots of bandwidth to look at this

recipe/meta.yaml
jakirkham: After attempts to remove `libgomp` on Windows, we have concluded it is needed here as a workaround. Ideally it should be added via `run_exports/strong`, but that is not happening for some reason, which will need to be investigated upstream in the toolchain. For now we will keep this here

Leaving this thread unresolved so the context is visible when searching back through here in the future

danielnachun:
Thanks, I'll fix the remaining `libgomp` issues.  Unfortunately it seems the CUDA R package on Windows is still failing - I think rather than trying to disable the package altogether I will try to just skip the test on the platform.

danielnachun:
@jakirkham I think we are ready to merge here!

recipe/meta.yaml
```diff
@@ -66,10 +68,11 @@ outputs:
         - cmake
         - make
         - llvm-openmp  # [osx]
-        - libgomp      # [linux]
+        - libgomp      # [linux or win]
       host:
         - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler != "None"]
         - nccl                                      # [linux and cuda_compiler != "None"]
+        - libgomp                                   # [win]
       run:
```
jakirkham, line 76: Think we need this in `run` as well

```suggestion
        - libgomp                                   # [win]
      run:
        - libgomp                                   # [win]
```

danielnachun: This is now fixed!  Hopefully we're ready to go now?

recipe/meta.yaml
```diff
@@ -258,11 +265,11 @@ outputs:
       files:
         - test-r-xgboost.r
       commands:
-        - Rscript test-r-xgboost.r
+        - Rscript test-r-xgboost.r  # [cuda_compiler == "None" or not win]
```
jakirkham, line 268: Think this is equivalent. This just makes it a bit clearer that the issue is Windows CUDA tests (likely due to the lack of GPU)

```suggestion
        - Rscript test-r-xgboost.r  # [not (win and cuda_compiler != "None")]
```

recipe/meta.yaml
```diff
@@ -66,11 +68,13 @@ outputs:
...
+        - libgomp      # [linux or win]
       host:
         - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler != "None"]
         - nccl                                      # [linux and cuda_compiler != "None"]
+        - libgomp                                   # [win]
       run:
+        - libgomp                                   # [win]
         - __cuda                                    # [cuda_compiler != "None"]
         - cuda-version >=11.2,<12.0a0               # [(cuda_compiler_version or "").startswith("11")]
         - cuda-version >=12.0,<13.0a0               # [(cuda_compiler_version or "").startswith("12")]
```
jakirkham, line 80: Swapping the order to match the sections above

```suggestion
      run:
        - __cuda                                    # [cuda_compiler != "None"]
        - cuda-version >=11.2,<12.0a0               # [(cuda_compiler_version or "").startswith("11")]
        - cuda-version >=12.0,<13.0a0               # [(cuda_compiler_version or "").startswith("12")]
        - libgomp                                   # [win]
```

recipe/meta.yaml
```diff
@@ -213,28 +217,30 @@ outputs:
...
+        - r-base                           # [build_platform != target_platform]
+        - r-matrix                         # [build_platform != target_platform]
+        - r-data.table                     # [build_platform != target_platform]
+        - r-magrittr                       # [build_platform != target_platform]
+        - r-jsonlite                       # [build_platform != target_platform]
+        - r-knitr                          # [build_platform != target_platform]
       host:
         - {{ pin_subpackage('_r-xgboost-mutex', exact=True) }}
         - {{ pin_subpackage('libxgboost', max_pin='x.x.x') }}
+        - libgomp      # [linux or win]
```
jakirkham, line 243: Tweak selector depth to make this easier to pick out

```suggestion
        - libgomp                                                  # [linux or win]
```

recipe/meta.yaml
```diff
@@ -245,6 +251,7 @@ outputs:
       run:
         - {{ pin_subpackage('_r-xgboost-mutex', exact=True) }}
         - {{ pin_subpackage('libxgboost', max_pin='x.x.x') }}
+        - libgomp  # [win]
```
jakirkham, line 254: Tweak selector depth to make this easier to pick out

```suggestion
        - libgomp                                                  # [linux or win]
```

recipe/meta.yaml
```diff
@@ -258,11 +265,11 @@ outputs:
       files:
         - test-r-xgboost.r
       commands:
-        - Rscript test-r-xgboost.r
+        - Rscript test-r-xgboost.r  # [not (win and cuda_compiler != "None")]
```
jakirkham, line 268: To track this, have filed issue: https://github.com/conda-forge/xgboost-feedstock/issues/205

jakirkham (approved):
Thanks Daniel for all your hard work here! 🙏

Also thanks everyone who helped review, debug, fix odd build issues, test failures, etc.. Truly a lot of great work went into this  👏

Let's get this in :shipit:

PR #279, opened 2026-02-26
Remove CCCL version pin

chyunsu3:
Checklist
* [x] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [x] Bumped the build number (if the version is unchanged)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

Closes #266

PR #301, opened 2026-08-07
[Don't merge] Bump 3.4.0

trivialfis:
Checklist

# Changes
- Bump XGBoost version.
- Cleanup/rebase patches.
- Implement https://github.com/conda-forge/xgboost-feedstock/pull/288#issuecomment-4816968208


trivialfis:

trivialfis:
@chyunsu3 I'm not familiar with the scikit build core, using `--config-settings=wheel.cmake=false` helps avoid the dependency on cmake/ninja/make for the Python package, but I'm not sure about its implications. Is the `use system xgboost` option still useful, or maybe we can remove it from XGBoost?

I opened the PR to check how many patches are still needed here. Feel free to take over if you have better insight, and I will close this PR.

chyunsu3:
Let's wait on this until https://github.com/conda-forge/xgboost-feedstock/pull/279 is merged.

recipe/meta.yaml
```diff
@@ -11,12 +11,10 @@ package:
 
 source:
   url: https://github.com/dmlc/xgboost/releases/download/v{{ version }}/xgboost-src-{{ version }}.tar.gz
-  sha256: 22d4fba822fba5cd02299bf0c63ec68ff72606bc1b1bd910423d4b83c2f108ff
+  sha256: af4588b34c7fa1bfde258006beebbe181454f1fb74266f81883b23059af3b9fb
   patches:
     # xgboost patches
     - patches/0001-Enable-latest-libcxx-on-MacOS.patch                # [osx]
-    - patches/0002-Remove-nvidia-nccl-cu12-from-pyproject.toml.patch
-    - patches/0003-Mark-wheels-as-any-platform-compatible.patch
```
jakirkham, line 19: Nice to see these two can be removed

recipe/install-py-xgboost.sh
```diff
@@ -3,6 +3,9 @@
 
 pushd ${SRC_DIR}/python-package
 
-    ${PYTHON} -m pip install . -vv --config-settings use_system_libxgboost=True
+    # Remove NCCL dependency
+    ${PYTHON} ${SRC_DIR}/ops/script/pypi_variants.py --use-suffix=na --require-nccl-dep=na
+    ${PYTHON} -m pip install . -vv \
+        --config-settings=wheel.platlib=false --config-settings=wheel.cmake=false
```
jakirkham: Let's increase the verbosity a bit so we have a better idea of what is happening with `scikit-build-core` to make sure it is as expected

Also we can break these into one per line to keep diffs clean if/when they change

```suggestion
    ${PYTHON} -m pip install . -vvv \
        --config-settings=wheel.platlib=false \
        --config-settings=wheel.cmake=false
```

chyunsu3:
@jakirkham Can you review?

recipe/meta.yaml
```diff
@@ -98,7 +96,7 @@ outputs:
         - {{ pin_subpackage('libxgboost', max_pin='x.x.x') }}
         - libxgboost * {{ string_prefix }}_h*_{{ PKG_BUILDNUM }}
         - python {{ python_min }}
-        - hatchling >=1.12.1
+        - scikit-build-core >=0.11.0
```
jakirkham, line 99: Great to see this change. Will need to make a few changes to `requirements/build` to accommodate this.

recipe/meta.yaml
jakirkham, line 94: Now that we are using `scikit-build-core`, we will need to add `cmake` and `ninja` here

```suggestion
      build:                                                   # [build_platform != target_platform]
        - python {{ python_min }}                              # [build_platform != target_platform]
        - cross-python_{{ target_platform }} {{ python_min }}  # [build_platform != target_platform]
        - cmake
        - ninja
```

jakirkham, line 91: This configures CMake to build with Ninja

```suggestion
    build:
      noarch: python
      string: {{ string_prefix }}_pyh{{ PKG_HASH }}_{{ PKG_BUILDNUM }}
      script_env:
        - CMAKE_GENERATOR=Ninja
```

recipe/meta.yaml
jakirkham: Somewhat tangential, but I noticed we still had one reference to `hatchling` upstream: https://github.com/dmlc/xgboost/pull/12219#discussion_r3752969374

recipe/meta.yaml
chyunsu3: I don't think so, since we are passing `wheel.cmake=false` so that CMake/Ninja is not a dependency of `py-xgboost`. See https://github.com/conda-forge/xgboost-feedstock/pull/301#issuecomment-5220981366. The native library is built as part of `libxgboost`, and `py-xgboost` will reuse the native library.

chyunsu3: See my comment above.

trivialfis:
I will send a PR upstream to remove the redundant option.

jakirkham:
In the future please wait until I've had a chance to respond before merging

recipe/meta.yaml
jakirkham: Good to know. AFAICT it doesn't appear to be downloading CMake or Ninja during the build. So this looks ok

